PR #10, opened 2021-12-06
support custom resource

Iceber:
When collecting custom resources, you need to specify the version to be synchronized.
eg.
```yaml
    resources:
    - group: apps
      resources:
      - deployments
    - group: cert-manager.io
      resources:
      - issuers
      versions:
      - v1
      - v1beta1
      - v1alpha2
```

If no version is specified, then the top three versions of the priority order in the member cluster will be synchronized by default.
```yaml
  status:
    resources:
    - group: cert-manager.io
      resources:
      - kind: Issuer
        namespaced: true
        resource: issuers
        syncConditions:
        - lastTransitionTime: "2021-12-07T02:14:21Z"
          status: Syncing
          storageVersion: v1
          version: v1
        - lastTransitionTime: "2021-12-07T03:57:08Z"
          status: Syncing
          storageVersion: v1beta1
          version: v1beta1
        - lastTransitionTime: "2021-12-07T03:57:08Z"
          status: Syncing
          storageVersion: v1alpha3
          version: v1alpha3
```


lou-lan (approved):

Iceber:
Since different versions may exist in different clusters, and the conversion strategy between these versions may also be `Webhook`, redundant data is saved if multiple versions are synchronized.

In the next stage, the versions of custom resources in different clusters are reconciled in the `clustersynchro manager` by watching for the `CRD` resources.

If there is an intersection between the versions of custom resources between clusters and the conversion strategy is `None, then only one version needs to be synchronized in the cluster and saved as the same version.

if the conversion strategy for a custom resource between clusters is `Webhook`, then it falls back to the [default synchronization policy](https://github.com/clusterpedia-io/clusterpedia/pull/10#issue-1071840038)

Issue #5, opened 2021-12-02
support setting the last `resource version` in the reflector


Iceber:
Support setting the resource version in the reflector before resource synchronizatiion.

In the weak network environment, the start and stop of `resource synchronization` may be more frequent, 
so setting the last `resource version` can avoid the full list

Iceber:
We need to refactor the reflector to use watch and then list to prevent the list from getting the full amount of resources in case of errors

PR #18, opened 2021-12-08
[WIP] support setting the last resource version in the reflector

Iceber:
https://github.com/clusterpedia-io/clusterpedia/issues/5

The `clustersynchro` will automatically stop the synchronization of resources after two consecutive failed cluster health checks, and resume synchronization when the cluster is healthy.

In weak network environments, this may happen more frequently, and to avoid pulling all the data each time you start a resource sync, we need to support setting the reflector's field


Iceber:
Need to refactor the `reflector`,I'm working on this.

Issue #4, opened 2021-12-02
support for pruning `manageFields` and `last-applied-configuration` fields


Iceber:
`Server-Side Apply` stores field management in the `manageFields`, which may not be useful during resource search.
This field needs to be pruned according to the user's needs.

The ` kubectl.kubernetes.io/last-applied-configuration` annotation also requires a feature gate to be provide  to prune.
```yaml
kind: Deployment
metadata:
  annotations:
    deployment.kubernetes.io/revision: "5"
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"apps/v1","kind":"Deployment","metadata":{"annotations":{},"labels":{"app":"cluster-autoscaler"},"name":"poc-autoscaler","namespace":"default"},"spec":{"replicas":1,"selector":{"matchLabels":{"app":"cluster-autoscaler"}},"template":{"metadata":{"labels":{"app":"cluster-autoscaler"}},"spec":{"containers":[{"args":["--cloud-provider=clusterapi","--kubeconfig=/etc/kubeconfig/value","--clusterapi-cloud-config-authoritative","--node-group-auto-discovery=clusterapi:clusterName=poc"],"command":["/cluster-autoscaler"],"image":"daocloud.io/daocloud/cluster-autoscaler:v1.20.1","name":"cluster-autoscaler","volumeMounts":[{"mountPath":"/etc/kubeconfig","name":"kubeconfig"}]}],"serviceAccountName":"cluster-autoscaler-management","terminationGracePeriodSeconds":10,"tolerations":[{"effect":"NoSchedule","key":"node-role.kubernetes.io/master"}],"volumes":[{"name":"kubeconfig","secret":{"secretName":"poc-kubeconfig"}}]}}}}
```

PR #19, opened 2021-12-09
prune managedFields and last-applied-configuration

Iceber:
The `clustersynchro` prunes fields based on `feature gate` when storing resources.
Pruning `managedFields` and `last-applied-configuration` annotation are supported.

```sh
./bin/clustersynchro-manager --feature-gates PruneManagedFields=true,PruneLastAppliedConfiguration=true 
```

Since the managedFields field is typically used very infrequently in resource searches, pruning prior to storage can also reduce the strain on storage and data transfer.

However, this may result in inconsistencies between the stored resources and the resource data in the member clusters.

issue: https://github.com/clusterpedia-io/clusterpedia/issues/4


Iceber:
Considering whether the `clusterpedia apiserver` should also support pruning to ensure that the complete data is stored and then pruned when the data is returned

PR #27, opened 2021-12-16
Allow CAData to be null and explicit Data is base64 encoded

wzshiming:


Iceber:
[examples/pediacluster.yaml](https://github.com/clusterpedia-io/clusterpedia/blob/main/examples/pediacluster.yaml) also needs to be updated.
```go
apiVersion: clusters.clusterpedia.io/v1alpha1
kind: PediaCluster
metadata:
  name: cluster-example
spec:
  apiserverURL: "https://10.30.43.43:6443"
  caData:
  tokenData:
  certData:
  keyData:
  resources:
  - group: apps
    resources:
     - deployments
  - group: ""
    resources:
     - pods
```

Iceber (approved):
LGTM

PR #29, opened 2021-12-20
update readme and pediacluster example

Iceber:
update readme and *pediacluster* example https://github.com/clusterpedia-io/clusterpedia/pull/27

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,10 +149,10 @@ metadata:
   name: cluster-example
 spec:
   apiserverURL: "https://10.30.43.43:6443"
-  caData: ""
-  tokenData: ""
-  certData: ""
-  keyData: ""
+  caData:
+  tokenData:
+  certData:
+  keyData:
   resources:
   - group: apps
     resources:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -149,10 +149,10 @@ metadata:
   name: cluster-example
 spec:
   apiserverURL: "https://10.30.43.43:6443"
-  caData: ""
-  tokenData: ""
-  certData: ""
-  keyData: ""
+  caData:
```
wzshiming, line 152: There is no need to remove this, `""` is still valid []byte

Iceber: but I got an error:
```
# pediaclusters.clusters.clusterpedia.io "cluster-x" was not valid:
# * spec.tokenData: Invalid value: "": spec.tokenData in body must be of type byte: ""
```

wzshiming: Well, I thought this validation rule was the same as JSON of Golang

wzshiming: LGTM

PR #33, opened 2021-12-21
support for the List object with remaining count and continue

Iceber:
add search label `search.clusterpedia.io/with-remaining-count` and `search.clusterpedia.io/with-continue`
|label|url query
|---|---|
|-selector  "`search.clusterpedia.io/with-remaining-count`=true"|`withRemainingCount`=true|
|--selector  "`search.clusterpedia.io/with-continue`=true"|`withContinue`=true|

example:
```bash
curl http://127.0.0.1:8080/api/v1/namespaces/default/pods?labelSelector=search.clusterpedia.io/with-remaining-count=true&limit=1&withContinue=true
{
    "kind": "PodList",
    "apiVersion": "v1",
    "metadata": {
        "continue": "1",
        "remainingItemCount": 1
    },
    "items": [
    ....
    ]
}
```

### DIFF
--- a/pkg/apis/pedia/v1alpha1/zz_generated.conversion.go
+++ b/pkg/apis/pedia/v1alpha1/zz_generated.conversion.go
@@ -193,6 +193,8 @@ func autoConvert_v1alpha1_ListOptions_To_pedia_ListOptions(in *ListOptions, out 
 	// WARNING: in.ClusterNames requires manual conversion: inconvertible types (string vs []string)
 	// WARNING: in.Namespaces requires manual conversion: inconvertible types (string vs []string)
 	// WARNING: in.OrderBy requires manual conversion: inconvertible types (string vs []github.com/clusterpedia-io/clusterpedia/pkg/apis/pedia.OrderBy)
+	out.WithContinue = (*bool)(unsafe.Pointer(in.WithContinue))
+	out.WithRemainingCount = (*bool)(unsafe.Pointer(in.WithRemainingCount))
 	return nil
 }
 
@@ -210,6 +212,8 @@ func autoConvert_pedia_ListOptions_To_v1alpha1_ListOptions(in *pedia.ListOptions
 		return err
 	}
 	// WARNING: in.OrderBy requires manual conversion: inconvertible types ([]github.com/clusterpedia-io/clusterpedia/pkg/apis/pedia.OrderBy vs string)
+	out.WithContinue = (*bool)(unsafe.Pointer(in.WithContinue))
+	out.WithRemainingCount = (*bool)(unsafe.Pointer(in.WithRemainingCount))
 	// WARNING: in.ExtraLabelSelector requires manual conversion: does not exist in peer-type
 	// WARNING: in.ExtraQuery requires manual conversion: does not exist in peer-type
 	return nil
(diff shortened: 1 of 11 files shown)


pkg/storage/internalstorage/util.go
```diff
@@ -83,11 +83,20 @@ func applyListOptionsToQuery(query *gorm.DB, opts *pediainternal.ListOptions) *g
...
 			}
 			query = query.Where(jsonQuery)
 		}
 	}
 
+	var amount *int64
+	if opts.AttachRemainingCount != nil && *opts.AttachRemainingCount {
+		var count int64
+		amount = &count
+		query = query.Count(amount)
```
wzshiming, line 97:
```suggestion
		query = query.Count(&amount)
```

pkg/apis/pedia/v1alpha1/zz_generated.conversion.go
```diff
@@ -210,6 +212,8 @@ func autoConvert_pedia_ListOptions_To_v1alpha1_ListOptions(in *pedia.ListOptions
 		return err
 	}
 	// WARNING: in.OrderBy requires manual conversion: inconvertible types ([]github.com/clusterpedia-io/clusterpedia/pkg/apis/pedia.OrderBy vs string)
```
wzshiming, line 214: There are also warnings, should be converted manually

pkg/storage/internalstorage/collectionresource_storage.go
```diff
@@ -39,7 +39,10 @@ func (s *CollectionResourceStorage) Get(ctx context.Context, opts *pediainternal
 
 		types[rt.GroupResource()] = &cr.ResourceTypes[i]
 	}
-	query = applyListOptionsToQuery(query, opts)
+
+	// collection resources don't support attach remaining count
+	opts.AttachRemainingCount = nil
```
wzshiming, line 44: It is not good to change the original data when using, whether to add a unified check

pkg/storage/internalstorage/util.go
Iceber: amount is a pointer

wzshiming: Yes, pointer should work.

pkg/apis/pedia/v1alpha1/zz_generated.conversion.go
Iceber: `autoConvert_pedia_ListOptions_To_v1alpha1_ListOptions` has been replaced with `Convert_pedia_ListOptions_To_v1alpha1_ListOptions` in`conversion.go`.
No suitable way found to prevent generating `autoConvert_pedia_ListOptions_To_v1alpha1_ListOptions`

pkg/storage/internalstorage/util.go
wzshiming:
``` golang
	var amount = new(int64)
	query = query.Count(amount)
```
Or
``` golang
	var amount int64
	query = query.Count(&amount)
```

Iceber: amount needs to be returned as a pointer, nil and 0 have different meanings

pkg/apiserver/registry/pedia/collectionresources/rest.go
```diff
@@ -88,6 +88,10 @@ func (s *REST) Get(ctx context.Context, name string, _ *metav1.GetOptions) (runt
 	query := request.RequestQueryFrom(ctx)
 	pediascheme.ParameterCodec.DecodeParameters(query, pediav1alpha1.SchemeGroupVersion, &opts)
 
+	// collection resources don't support attach remaining count
+	// ignore opts.AttachRemainingCount
+	opts.AttachRemainingCount = nil
```
Iceber, line 93: @wzshiming  The ignore operation is placed where the option is generated, and support will be provided in the future

Iceber:
Does not support List as ParitalObjectMetadataList with count, other cases will come with count.
https://github.com/kubernetes/kubernetes/pull/107177

wzshiming:
LGTM

wzshiming (approved):

Iceber:
After a simple discussion, it was decided to use `WithRemaningCount` and `WithContinue`

PR #36, opened 2021-12-28
support field selector

Iceber:
For `fieldSelector`, kubernetes currently only supports `metadata.name` and `metadata.namespace`,and the only operations supported are `=`, `!=`, `==`, which is a bit weak.

clusterpedia has enhanced its support for `fieldSelector`, and supports the same operations as `labelSelector`: `!`, `=` ,`!=`, `==`, `in`, `not in`

examples:
```bash
kubectl get pods --field-selector="status.phase=Running"

kubectl get pods --field-selector=".status.phase=Running"
```
Available whether or not it starts with `.`.

If the field contains the characters `.`, then you need to use `''`, `""` or `[' ']` to wrap
```bash
kubectl get deploy --field-selector="metadata.annotations['test.io'] in (value1,value2),spec.replica=3"

kubectl get deploy --field-selector="metadata.annotations.'test.io' in (value1,value2),spec.replica=3"
```

There is also support for list fields:
```bash
kubectl get po --field-selector="spec.containers[].name!=container1"

kubectl get po --field-selector="spec.containers[].name == container1"

kubectl get po --field-selector="spec.containers[1].name in (container1,container2)"
```
**But the `internalstorage` does not support list field filtering at this time.**

### DIFF
--- a/pkg/apis/pedia/types.go
+++ b/pkg/apis/pedia/types.go
@@ -8,6 +8,8 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/clusterpedia-io/clusterpedia/pkg/utils/fields"
 )
 
 const (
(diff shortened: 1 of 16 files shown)


wzshiming (approved):
/lgtm

PR #37, opened 2021-12-29
Move the cluster name of resources to annotations

calvin0327:
It's more appropriate to put the cluster message on annotions.

### DIFF
--- a/pkg/utils/clustername.go
+++ b/pkg/utils/clustername.go
@@ -9,8 +9,8 @@ import (
 
 func ExtractClusterName(obj runtime.Object) string {
 	if m, err := meta.Accessor(obj); err == nil {
-		if labels := m.GetLabels(); labels != nil {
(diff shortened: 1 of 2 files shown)


pkg/utils/clustername.go
```diff
@@ -1,6 +1,7 @@
 package utils
 
 import (
+	"google.golang.org/genproto/googleapis/api/annotations"
```
Iceber, line 4: Looks like unnecessary packages were imported.

calvin0327: Let me clean up it.

Iceber:
This modification will break the synchronized resources and the user needs to clean up the synchronized resource data.
> Currently in alpha stage, no need for compatibility processing
```sql
DROP TABLE resources;
```

Iceber (approved):
LGTM

PR #38, opened 2021-12-29
Update the readme

calvin0327:
The readme have a  label typo `search.clsuterpedia.io/offset`.


Iceber:
In the readme, the `url query` is also wrong.
|desc|label|url query|example|
|---|---|---|---|
|Specified size |search.clusterpedia.io/size|limit|`?limit=100`
|Specified offset |search.clusterpedia.io/offset|continue|`?continue=10`

Which do you think is better, `search.clusterpedia.io/size` or `search.clusterpedia.io/limit`?

calvin0327:
I think the label `search.clusterpedia.io/limit` is better.

In general,  `offsize` and `limit` come in pairs to make it easier to understand.

Iceber (approved):
LGTM

PR #42, opened 2022-01-06
change size label to limit label.

calvin0327:
Change the label `size` to `limit`, I think `limit` is  easier to understand.


Iceber (approved):
LGTM

PR #44, opened 2022-01-06
internalstorage: do not set the default order by

Iceber:
Due to performance reasons, the default order by is not set.
The user needs to use the search label `search.clusterpedia.io/orderby` to actively set the sorted fields.

```bash
kubectl --cluster clusterpedia get deployments -A -l "search.clusterpedia.io/orderby in (cluster, namespace, name)"
```


calvin0327:
LGTM

PR #46, opened 2022-01-07
fix update codegen script

scydas:


Iceber:
Thank you for finding the issue!

Issue #43, opened 2022-01-06
clusterpedia required helm deployment


calvin0327:
* deploy api server
* deploy clustersynchro manager
* deploy a usable database, `mysql` or `postgresql`
* relevant `RBAC` and serviceAccout

PR #53, opened 2022-01-19
use helm build clusterpedia

calvin0327:
fixed: https://github.com/clusterpedia-io/clusterpedia/issues/43

### DIFF
--- a/charts/templates/apiserver-apiservice.yaml
+++ b/charts/templates/apiserver-apiservice.yaml
@@ -1,0 +1,14 @@
+apiVersion: apiregistration.k8s.io/v1
+kind: APIService
+metadata:
+  name: v1beta1.clusterpedia.io
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+spec:
+  insecureSkipTLSVerify: true
+  group: clusterpedia.io
+  groupPriorityMinimum: 1000
+  versionPriority: 100
+  service:
+    name: {{ include "clusterpedia.apiserver.fullname" . }}
+    namespace: {{ .Release.Namespace }}
+  version: v1beta1
--- a/charts/templates/apiserver-deployment.yaml
+++ b/charts/templates/apiserver-deployment.yaml
@@ -1,0 +1,67 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "clusterpedia.apiserver.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app: {{ include "clusterpedia.apiserver.fullname" . }}
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+spec:
+  replicas: {{ .Values.apiserver.replicaCount }}
+  selector:
+    matchLabels: {{- include "common.labels.matchLabels" . | nindent 6 }}
+      app: {{ include "clusterpedia.apiserver.fullname" . }}
+  template:
+    metadata:
+      {{- if .Values.apiserver.podAnnotations }}
+      annotations:
+      {{- include "common.tplvalues.render" (dict "value" .Values.apiserver.podAnnotations "context" $) | nindent 8 }}
+      {{- end }}
+      labels: {{- include "common.labels.standard" . | nindent 8 }}
+        app: {{ include "clusterpedia.apiserver.fullname" . }}
+        {{- if .Values.podLabels }}
+        {{- include "common.tplvalues.render" (dict "value" .Values.apiserver.podLabels "context" $) | nindent 8 }}
+        {{- end }}
+    spec:
+      {{- include "clusterpedia.apiserver.imagePullSecrets" . | nindent 6 }}
+      containers:
+      - name: {{ include "clusterpedia.apiserver.fullname" . }}
+        image: {{ template "clusterpedia.apiserver.image" . }}
+        imagePullPolicy: {{ .Values.apiserver.image.pullPolicy }}
+        command:
+        - /usr/local/bin/apiserver
+        - --secure-port=443
+        - --storage-config=/etc/clusterpedia/storage/internalstorage-config.yaml
+        - -v=3
+        {{- if .Values.apiserver.resources }}
+        resources: {{- toYaml .Values.apiserver.resources | nindent 12 }}
+        {{- end }}
+        env:
+        - name: DB_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              name: {{ include "clusterpedia.internalstorage.fullname" . }}
+              key: password
+        volumeMounts:
+        - name: internalstorage-config
+          mountPath: /etc/clusterpedia/storage
+          readOnly: true
+      serviceAccountName: {{ include "clusterpedia.apiserver.fullname" . }}
+      {{- if .Values.apiserver.affinity }}
+      affinity: {{- include "common.tplvalues.render" (dict "value" .Values.apiserver.affinity "context" $) | nindent 8 }}
+      {{- end }}
+      {{- if .Values.apiserver.nodeSelector }}
+      nodeSelector: {{- include "common.tplvalues.render" (dict "value" .Values.apiserver.nodeSelector "context" $) | nindent 8 }}
+      {{- end }}
+      {{- if .Values.apiserver.tolerations }}
+      tolerations: {{- include "common.tplvalues.render" (dict "value" .Values.apiserver.tolerations "context" $) | nindent 8 }}
+      {{- end }}
+      volumes:
+        - name: internalstorage-config
+          configMap:
+            name: {{ include "clusterpedia.internalstorage.fullname" . }}
--- a/charts/templates/clustersynchro-manager-deployment.yaml
+++ b/charts/templates/clustersynchro-manager-deployment.yaml
@@ -1,0 +1,67 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "clusterpedia.clustersynchroManager.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app: {{ include "clusterpedia.clustersynchroManager.fullname" . }}
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+spec:
+  replicas: {{  .Values.clustersynchroManager.replicaCount }}
+  selector:
+    matchLabels: {{- include "common.labels.matchLabels" . | nindent 6 }}
+      app: {{ include "clusterpedia.clustersynchroManager.fullname" . }}
+  template:
+    metadata:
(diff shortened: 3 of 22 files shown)


charts/charts/mysql/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v2
```
Iceber, line 1: Could We use the existing mysql charts，example https://artifacthub.io/packages/helm/bitnami/mysql

calvin0327:
@Iceber  PTAL update the internal storage dependecy to bitnami repo and default use the postgresql as our storage.

charts/templates/internalstorage-configmap.yaml
```diff
@@ -0,0 +1,18 @@
...
+metadata:
+  name: {{ $name }}-internalstorage
+  namespace: {{ include "clusterpedia.namespace" . }}
+data:
+  internalstorage-config.yaml: |-
+    type: {{- include "clusterpedia.storage.type" . | quote | nindent 6 }}
+    host: {{- include "clusterpedia.storage.host" . | quote | nindent 6 }}
+    port: {{- include "clusterpedia.storage.port" . | nindent 6 }}
+    user: {{- include "clusterpedia.storage.user" . | nindent 6 }}
+    database: "clusterpedia"
```
Iceber, line 14: `helm template` ->
```yaml
# Source: clusterpedia/templates/internalstorage-configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: quickstart-internalstorage
  namespace: default
data:
  internalstorage-config.yaml: |-
    type:
      "postgres"
    host:
      "quickstart-postgresql"
    port:
      5432
    user:
      postgres
    database: "clusterpedia"
```

fix:
```yaml
  internalstorage-config.yaml: |-
    type: {{ include "clusterpedia.storage.type" . | quote }}
    host: {{ include "clusterpedia.storage.host" . | quote }}
    port: {{ include "clusterpedia.storage.port" . }}
    user: {{ include "clusterpedia.storage.user" .  | quote}}
```

charts/templates/apiserver-apiservice.yaml
```diff
@@ -0,0 +1,16 @@
+{{- $name := include "clusterpedia.name" . -}}
```
Iceber, line 1: I think that declaring template variables again in a separate file does not do any good and adds a degree of complexity.

charts/templates/crds/clusters.clusterpedia.io_pediaclusters.yaml
```diff
@@ -0,0 +1,222 @@
+
```
Iceber, line 1: For the management of crds in charts, can we look at cert-manager or is there a better solution?
ref: 
* https://cert-manager.io/docs/installation/helm/#option-2-install-crds-as-part-of-the-helm-release
* https://cert-manager.io/docs/installation/upgrading/#upgrading-with-helm

charts/values.yaml
```diff
@@ -0,0 +1,238 @@
...
+    ##
+    selector: {}
+  ## @section MySQL Primary parameters
+  primary:
+    ## MySQL Primary Service parameters
+    ##
+    service:
+      ## @param primary.service.port MySQL Primary K8s service port
+      ##
+      port: 3306
```
Iceber, line 238: If the values of the subcharts are the same as the default values can they be ignored in values.yaml to avoid too much complexity?

charts/templates/apiserver-apiservice.yaml
calvin0327: For a separate file that only needs to use the value once, this really doesn't make sense, but for a file that uses the value multiple times, it's much more readable.

charts/templates/crds/clusters.clusterpedia.io_pediaclusters.yaml
calvin0327: I looked at the cert-manager, the flag `installCRDs` is used to determine whether or not to install CRD in updation.

This a useful flag to us.

calvin0327:
@Iceber PTAL have amended above problem.

charts/values.yaml
```diff
@@ -0,0 +1,250 @@
...
+    # requests:
+    #   cpu: 100m
+    #   memory: 128Mi
+  ## @param apiServer.tolerations
+  tolerations:
+     - key: node-role.kubernetes.io/master
+       operator: Exists
+
+## controller manager config
+controllerManager:
```
Iceber, line 67: We don't have `controller manager`, only `clustersynchro manager` for now

charts/templates/apiserver-deployment.yaml
```diff
@@ -0,0 +1,58 @@
...
+    metadata:
+      {{- with .Values.apiServer.podAnnotations}}
+      annotations:
+      {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+      {{- include "clusterpedia.apiserver.labels" . | nindent 8}}
+      {{- include "clusterpedia.apiserver.podLabels" . | nindent 8}}
+    spec:
+      {{- with .Values.controllerManager.imagePullSecrets}}
```
Iceber, line 25: `.Values.apiserver.imagePullSecrets` ?

calvin0327:
@Iceber /cc

charts/values.yaml
```diff
@@ -0,0 +1,250 @@
...
+    ##
+    selector: {}
+  ## @section MySQL Primary parameters
+  primary:
+    ## MySQL Primary Service parameters
+    ##
+    service:
+      ## @param primary.service.port MySQL Primary K8s service port
+      ##
+      port:
```
Iceber, line 250: I think for the subchart values, only the override fields that are retained will be fine.
```yaml

## @section PostgreSQL Parameters
##
## PostgreSQL properties
##
postgresql:
  ## @param postgresql.enabled Deploy the PostgreSQL sub-chart
  ##
  enabled: true
  auth:
    ## @param auth.postgresPassword Password for the "postgres" admin user
    ##
    postgresPassword: "dangerous0"
    ## @param postgresql.auth.database Postgresql database
    ##
    database: clusterpedia

  ## PostgreSQL Primary persistence configuration
  ##
  persistence:
    ## @param primary.persistence.enabled Enable PostgreSQL Primary data persistence using PVC
    ##
    enabled: true
    ## @param primary.persistence.size PVC Storage Request for PostgreSQL volume
    ##
    size: 10Gi

## @section MySQL Parameters
##
## MySQL properties
##
mysql:
  ## @param enabled Enable mysql
  enabled: false
  ## Properties for using an existing mysql installation
  ## @param commonLabels [object] Common labels to add to all MySQL resources (sub-charts are not considered). Evaluated as a template
  commonLabels:
    app: clusterpedia-internalstorage
    internalstorage.clusterpedia.io/type: mysql
  auth:
    ## @param auth.rootPassword Password for the `root` user. Ignored if existing secret is provided
    ##
    rootPassword: "dangerous0"
    ## @param auth.database Name for a custom database to create
    ##
    database: clusterpedia

  ## Enable persistence using Persistent Volume Claims
  ##
  persistence:
    ## @param primary.persistence.enabled Enable persistence on MySQL primary replicas using a `PersistentVolumeClaim`. If false, use emptyDir
    ##
    enabled: true
    ## @param primary.persistence.size MySQL primary persistent volume size
    ##
    size: 10Gi
```

charts/values.yaml
```diff
@@ -0,0 +1,250 @@
...
+  persistence:
+    ## @param global.persistence.matchNode define the nodeAffinity of pv.
+    ## if defined, will create the pv on the node
+    matchNode: kind-control-plane
+  ## @param global.postgresql.auth define the auth info.
+  postgresql:
+    auth:
+      ## @param auth.postgresPassword Password for the "postgres" admin user
+      ##
+      postgresPassword: "dangerous"
```
Iceber, line 22: Overriding `global.postgres` doesn't seem to make sense, `postgresql` will be overridden below

charts/values.yaml
```diff
@@ -0,0 +1,250 @@
...
+  ## @param dbInstallMode "internal" and "external" are provided
+  ## "internal" means automatic Database installation, default mysql
+  ## "external" means to install the database yourself, must specify the database connection parameters
+  storageInstallMode: "internal"
+  ## persistence configuration
+  ##
+  persistence:
+    ## @param global.persistence.matchNode define the nodeAffinity of pv.
+    ## if defined, will create the pv on the node
+    matchNode: kind-control-plane
```
Iceber, line 16: `storageInstallMode` and `persistence` don't seem to make sense for `global`, only for clusterepdia, there is no need to put them in `global`

Iceber: `imageRegistry` and `imagePullSecrets` look like they could be placed in global.
example:
```yaml
global:
## @param global.imageRegistry Global Docker image registry
##
imageRegistry: ""
## @param global.imagePullSecrets Global Docker registry secret names as an array
## e.g.
## imagePullSecrets:
##   - myRegistryKeySecretName
##
imagePullSecrets: []
```

charts/values.yaml
```diff
@@ -0,0 +1,250 @@
...
+  replicaCount: 1
+  ## @param apiServer.podAnnotations
+  podAnnotations: { }
+  ## @param apiServer.podLabels
+  podLabels: { }
+  ## @param apiServer.imagePullSecrets
+  imagePullSecrets: []
+  image:
+    ## @param apiServer.image.repository image of the apiserver
+    repository: ghcr.io/clusterpedia-io/clusterpedia/apiserver
```
Iceber, line 46: Splitting the `repository` into `registry` and `repository` seems to make it easier to use

charts/values.yaml
```diff
@@ -0,0 +1,250 @@
...
+database: postgresql
+## @param installCRDs define flag whether to install CRD resources
+## 
+installCRDs: false
+
+## clusterpedia apiserver config
+apiServer:
+  ## @param apiServer.labels
+  labels: 
+    app: clusterpedia-apiserver
```
Iceber, line 35: `app: clusterpedia-apiserver` as basic lables should be put in _helpers.tml, not here, otherwise it will be overwritten by the user

charts/templates/clustersynchro-manager-deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+      imagePullSecrets:
+      {{- toYaml . | nindent 8 }}
+      {{- end }}
+      containers:
+      - name: {{ $name }}-clustersynchro-manager
+        image: "{{ .Values.clustersynchroManager.image.repository }}:{{ .Values.clustersynchroManager.image.tag | default  "latest" }}"
+        command:
+        - /usr/local/bin/clustersynchro-manager
+        - --storage-config=/etc/clusterpedia/storage/internalstorage-config.yaml
+        - --feature-gates=PruneManagedFields=true,PruneLastAppliedConfiguration=true
```
Iceber, line 35: ` --leader-elect-resource-namespace` flag needs to be set, otherwise the leader election resource cannot be created becuase of the namespace change.

charts/templates/internalstorage-configmap.yaml
```diff
@@ -0,0 +1,17 @@
...
+  internalstorage-config.yaml: |-
+    type: {{ include "clusterpedia.storage.type" . | quote }}
+    host: {{ include "clusterpedia.storage.host" . | quote }}
+    port: {{ include "clusterpedia.storage.port" . }}
+    user: {{ include "clusterpedia.storage.user" . }}
+    database: "clusterpedia"
+    {{ if eq (include "clusterpedia.storage.type" .) "mysql" -}}
+    log:
+      slowThreshold: "100ms"
+    {{- end -}}
```
Iceber, line 17: We can put the log configuration in values.yaml

charts/templates/internalstorage-local-pv.yaml
```diff
@@ -0,0 +1,25 @@
...
+  name: {{ include "clusterpedia.name" . }}-internalstorage
+spec:
+  capacity:
+    storage: {{ include "clusterpedia.internalstorage.capacity" . }}
+  volumeMode: Filesystem
+  accessModes:
+  - ReadWriteOnce
+  persistentVolumeReclaimPolicy: Retain
+  local:
+    path: {{ include "clusterpedia.job.storage.hostPath" . }}
```
Iceber, line 14: clusterpedia.job.storage -> clusterpedia.storage ?

Iceber:
@calvin0327 Thank you very much for your contribution, using helm to deploy is a very important feature!

PR #55, opened 2022-01-26
internalstorage: fix update resource

Iceber:

The uid may not be the same for resources with the same namespace/name in the same cluster at different times.

For example, if a resource in the cluster is deleted during a long period of time without synchronization and a resource with the same name is created, the UID will not be the same.


scydas:
LGTM

PR #69, opened 2022-02-10
set the latest image version to "latest" 

Iceber:

The alpha development phase of clusterpedia is complete and ready for release `0.1.0` into the official version iteration

In the future, the latest image will be the `latest` version, and the image with tag will contain both `latest` and `tag` versions


calvin0327:
lgtm

PR #99, opened 2022-03-09
add tests for resources rest

Iceber:

### DIFF
--- a/pkg/apiserver/registry/clusterpedia/resources/rest_test.go
+++ b/pkg/apiserver/registry/clusterpedia/resources/rest_test.go
@@ -1,0 +1,155 @@
+package resources
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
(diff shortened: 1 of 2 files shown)


pkg/apiserver/registry/clusterpedia/resources/rest_test.go
```diff
@@ -0,0 +1,149 @@
...
+		url string
+
+		expectedCode int
+		expectedBody string
+	}{
+		{
+			"/api/v1/resources",
+
+			200,
+			newResponseBody("", ""),
```
Iceber, line 63: This doesn't look very intuitive, so we can use `expectedCluster` and `expectedURL` and generate the response body in the test

Iceber, line 63: This doesn't look very intuitive, so we can use `expectedCluster` and `expectedURL` and generate the response body in the test

PR #100, opened 2022-03-09
add tests for utils

Iceber:

### DIFF
--- a/pkg/utils/filters/query_test.go
+++ b/pkg/utils/filters/query_test.go
@@ -1,0 +1,105 @@
+package filters
+
+import (
+	"net/http"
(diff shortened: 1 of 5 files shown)


pkg/utils/filters/query_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/clusterpedia-io/clusterpedia/pkg/utils/request"
+)
+
+func TestWithRequestQuery(t *testing.T) {
+	tests := []struct {
+		name string
+		url  string
```
Iceber, line 16: Explicitly expected values may be easier to understand.

PR #106, opened 2022-03-11
internalstorage: add tests for resource storage

Iceber:


pkg/storage/internalstorage/resource_storage_test.go
```diff
@@ -0,0 +1,369 @@
...
+		t.Run(fmt.Sprintf("%s mysql-%s", name, version), func(t *testing.T) {
+			mysqlDB := mysqlDBs[version]
+			mysqlSQL, err := toSQL(mysqlDB, options, func(query *gorm.DB, options *internal.ListOptions) (*gorm.DB, error) {
+				_, _, query, err := applyListOptionsToResourceQuery(mysqlDB, query, options)
+				return query, err
+			},
+			)
+			if err != nil && err.Error() != expected.err {
+				t.Errorf("expected error: %q, but got: %q", err.Error(), expected.err)
+			}
```
Iceber, line 41: If `err == nil`, but the `expected.err != ""`, then the check is skipped.

pkg/storage/internalstorage/resource_storage_test.go
```diff
@@ -0,0 +1,369 @@
...
+)
+
+func testApplyListOptionsToResourceQuery(t *testing.T, name string, options *internal.ListOptions, expected expected) {
+	t.Run(fmt.Sprintf("%s postgres", name), func(t *testing.T) {
+		postgreSQL, err := toSQL(postgresDB, options, func(query *gorm.DB, options *internal.ListOptions) (*gorm.DB, error) {
+			_, _, query, err := applyListOptionsToResourceQuery(postgresDB, query, options)
+			return query, err
+		},
+		)
+		if err != nil && err.Error() != expected.err {
```
Iceber, line 22: If err == nil, but the expected.err != "", then the check is skipped.

pkg/storage/internalstorage/resource_storage_test.go
```diff
@@ -0,0 +1,369 @@
...
+		resource     schema.GroupVersionResource
+		cluster      string
+		namespace    string
+		resourceName string
+		expected     expected
+	}{
+		{
+			"empty",
+			appsv1.SchemeGroupVersion.WithResource("deployments"),
+			"",
```
Iceber, line 194: Need to test when gvk is empty

PR #110, opened 2022-03-12
add synced state to resource

calvin0327:
If the respurce had be synced by synchronizer， we should add a state  `Synced` to mark current state.

### DIFF
--- a/charts/_crds/cluster.clusterpedia.io_pediaclusters.yaml
+++ b/charts/_crds/cluster.clusterpedia.io_pediaclusters.yaml
@@ -58,6 +58,8 @@ spec:
               kubeconfig:
                 format: byte
                 type: string
+              syncAllCustomResources:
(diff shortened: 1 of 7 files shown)


calvin0327:
@Iceber PTAL

Iceber:
I think there are three types of sync statuses:
* [`init synced`] first sync completed
* [`syncing`] resource event received and processing started
* [`synced`] the current resource queue emptied

Normally, the resource state should switch between `synced` and `syncing`.

@calvin0327 This is the idea for me, what do you think?

calvin0327:
@Iceber I agree with that the resource state should switch between synced and syncing.  It's not necessary to add `init synced`  state, we can ues the `synced` replace it.

calvin0327:
@Iceber /cc

Iceber (approved):
LGTM

PR #113, opened 2022-03-14
add filter by creation time

cleverhu:
Two search tags and query keys are planned to be added: search.clusterpedia.io/since and search.clusterpedia.io/before.

The time range is closed at the front and open at the back([, )).

Three time formats will be supported for added convenience:

1、unix time（seconds/millseconds）                      supported by labels and query param(since&&before)
2、RFC3339(2006-01-02T15:04:05Z07:00)          supported by query param
3、UTC Date(2006-01-02)                                       supported by labels and query param
4、UTC Datetime(2006-01-02 15:04:05)                supported by query param

### DIFF
--- a/pkg/apis/clusterpedia/v1beta1/conversion.go
+++ b/pkg/apis/clusterpedia/v1beta1/conversion.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"time"
 
 	metainternal "k8s.io/apimachinery/pkg/apis/meta/internalversion"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -59,6 +60,14 @@ func Convert_v1beta1_ListOptions_To_clusterpedia_ListOptions(in *ListOptions, ou
 	}
 	out.OwnerSeniority = in.OwnerSeniority
 
+	if err := convert_String_To_Pointer_metav1_Time(&in.Since, &out.Since, nil); err != nil {
+		return err
+	}
+
+	if err := convert_String_To_Pointer_metav1_Time(&in.Before, &out.Before, nil); err != nil {
+		return err
+	}
+
 	out.WithContinue = in.WithContinue
 	out.WithRemainingCount = in.WithRemainingCount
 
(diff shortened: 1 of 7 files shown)


pkg/apis/clusterpedia/v1beta1/conversion.go
```diff
@@ -229,6 +252,45 @@ func convert_Slice_string_To_String(in *[]string, out *string, scope conversion.
...
+		}
+		if len(*in) == 10 {
+			t := time.Unix(timestamp, 0)
+			*out = &metav1.Time{Time: t}
+			return nil
+		} else if len(*in) == 13 {
+			t := time.UnixMilli(timestamp)
+			*out = &metav1.Time{Time: t}
+			return nil
+		}
```
Iceber, line 286: Distinguishing timestamp units by the length of the timestamp string is indeed a more user-friendly way of doing things, but there may be some unusual behaviour when used incorrectly by users, e.g. since is 10 and before is 13.

We need to determine whether before is greater than since.

pkg/apis/clusterpedia/v1beta1/conversion.go
```diff
@@ -229,6 +252,45 @@ func convert_Slice_string_To_String(in *[]string, out *string, scope conversion.
...
+	}
+	if *in != "" && out == nil {
+		return fmt.Errorf("Invalid time format(%s)", *in)
+	}
```
Iceber, line 290: `in` and `out` will not be nil in the conversion function

pkg/apis/clusterpedia/v1beta1/conversion.go
```diff
@@ -229,6 +252,45 @@ func convert_Slice_string_To_String(in *[]string, out *string, scope conversion.
...
+	return nil
+}
+
```
Iceber, line 293: Suggestions：
```go
func convert_String_To_Pointer_metav1_Time(in *string, out **metav1.Time, scope conversion.Scope) error {
        str := strings.TrimSpace(*in)
        if len(str) == 0 {
                return nil
        }

        var err error
        var t time.Time
        switch {
        case strings.Contains(str, " "):
                t, err = time.Parse("2006-01-02 15:04:05", *in)
        case strings.Contains(str, "T"):
                t, err = time.Parse(time.RFC3339, *in)
        case strings.Contains(str, "-"):
                t, err = time.Parse("2006-01-02", *in)
        default:
                var timestamp int64
                timestamp, err = strconv.ParseInt(*in, 10, 64)
                if err != nil {
                        break
                }

                switch len(str) {
                case 10:
                        t = time.Unix(timestamp, 0)
                case 13:
                        t = time.Unix(timestamp/1e3, (timestamp%1e3)*1e6)
                default:
                        return errors.New("Invalid timestamp: only timestamps with string lengths of 10(as s) and 13(as ms) are supported")
                }
        }
        if err != nil {
                return fmt.Errorf("Invalid datetime: %s, a valid datetime format: RFC3339, Datetime(2006-01-02 15:04:05), Date(2006-01-02), Unix Timestamp", err)
        }
        *out = &metav1.Time{Time: t}
        return nil
}
```

pkg/storage/internalstorage/util.go
```diff
@@ -47,6 +47,14 @@ func applyListOptionsToQuery(query *gorm.DB, opts *internal.ListOptions, applyFn
 		query = query.Where("name IN ?", opts.Names)
 	}
 
+	if opts.Since != nil {
+		query = query.Where("created_at >= ?", opts.Since.Time)
+	}
+
+	if opts.Before != nil {
+		query = query.Where("created_at < ?", opts.Before.Time)
```
Iceber, line 55: opts.Before.UTC()

pkg/apis/clusterpedia/v1beta1/conversion.go
```diff
@@ -159,6 +180,9 @@ func Convert_v1beta1_ListOptions_To_clusterpedia_ListOptions(in *ListOptions, ou
 			out.ExtraLabelSelector = labels.NewSelector().Add(extraLabelRequest...)
 		}
 	}
+	if out.Since != nil && out.Before != nil && out.Since.Sub(out.Before.Time).Seconds() > 0 {
```
Iceber, line 183: (*metav1.Time).Before can handle nil.
```go
if out.Since.Before(out.Before) {
}
```

Iceber:
@cleverhu LGTM, Thank you for your contribution.

PR #115, opened 2022-03-17
Feature: Support kubeconfig

### LINKED ISSUE #95: Support for cluster authentication via kubeconfig

Use kubeconfig directly, instead of filling in ca, cert, key, etc. separately in PediaCluster

### Why is this needed?

Using kubeconfig will make the creation of pediacluster much easier.

wzshiming:
Fix #95


pkg/apis/cluster/v1alpha2/types.go
```diff
@@ -32,9 +32,11 @@ type PediaCluster struct {
...
 
 type ClusterSpec struct {
-	// +required
-	// +kubebuilder:validation:Required
-	APIServer string `json:"apiserver"`
+	// +optional
+	Kubeconfig []byte `json:"kubeconfig,omitempty"`
+
+	// +optional
+	APIServer string `json:"apiserver,omitempty"`
```
Iceber, line 39: >// +kubebuilder:resource:scope="Cluster"
// +kubebuilder:printcolumn:name="APIServer",type=string,JSONPath=".spec.apiserver"
// +kubebuilder:printcolumn:name="Version",type=string,JSONPath=".status.version"
// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=".status.conditions[?(@.type == 'Ready')].reason"
`.spec.apiserver` will be used for the column. If the user does not fill in the apiserver, how can we display the apiserver address?

Add a field to the status to show the apiserver?

wzshiming: Maybe a Webhook should extract the Apiserver from KubeconFig and fill it back in

Iceber: MutationgWebhook is indeed a good way, but introducing additional components for `apiserver` field, of course, can be integrated into APIServer or ClusterSynchro Manager, but this also requires additional development work (in the future).

> I think CR name is more representative of a cluster.

`APIServer` printcolumn enables users to easily obtain the basic cluster information.

---
In the absence of webhook, is it necessary to force the user to additionally set the redundant apiserver field after setting up the kubeconfig? 
Or allow apiserver field to be empty and show *empty string* or *InKubeConfig* in printcolumn.
> In the future, webhook can be used to automatically parse the kubeconfig and then set the apiserver field.

@wzshiming @27149chen I'm not sure that's right, what do you think? 🤔

Iceber: After thinking about it again, I think that after setting kubeconfig, `apiserver` is not required, if the user needs to display in APIServer printcolumn, then set up redundant apiserver field.

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -304,6 +306,12 @@ func (manager *Manager) UpdateClusterStatus(ctx context.Context, name string, st
 }
 
 func buildClusterConfig(cluster *clusterv1alpha2.PediaCluster) (*rest.Config, error) {
+	if len(cluster.Spec.Kubeconfig) != 0 {
+		return clientcmd.BuildConfigFromKubeconfigGetter("", func() (conf *clientcmdapi.Config, err error) {
```
27149chen, line 310: I think you can use `clientcmd.NewClientConfigFromBytes()` here

pkg/apis/cluster/v1alpha2/types.go
27149chen: I agree with you that `apiserver ` is optional

pkg/synchromanager/clustersynchro_manager.go
wzshiming: make sense, updated.

Iceber (approved):
LGTM

27149chen (approved):

PR #119, opened 2022-03-26
separate apis from clusterpedia

27149chen:
This pr separates the api definitions from main clusterpedia module, changes:
1. move path pkg/apis to staging/api
2. make path staging/api as an individual module (github.com/clusterpedia-io/clusterpedia-api)

How to use the new module?
Please refer to staging/api/README.md for details


Iceber:
@27149chen Thank you very much for contributing!

There are a few small issues that may need to be discussed, mainly the directory where `apis` is placed, the package name and the name of the imported packages

kubernetes has adopted the `k8s.io/api` package name because it contains not only resources under */apis* but also resources under the path */api,* and for the sake of generality it might be called `k8s.io/api` (of course, I'm not sure why `k8s.io/api` is called `api`), but all our resources are under */apis*. Wouldn't it be better to call it `apis`?

In addition, it is common for staging to be structured in the same way as GOPATH. We should actually put it under `staging/src/github.com/clusterpedia-io/apis`. This style of directory also makes it easier to put the generated code under `staging` in the future.

For the imported package name - `github.com/clusterpedia.io/clusterpedia-api`, I feel that the clusterpedia prefix is not needed for `clusterepdia-api`, as the organization name actually indicates that the apis are under *clusterpedia.io*, so I suggest omitting the clusterpedia prefix.

These are my thoughts, what do you think?

27149chen:
@Iceber  thank you for your comments, here is my reply:

1. The reason why I name it api instead of apis is because it is a best practice in Go. See here: https://dmitri.shuralyov.com/idiomatic-go#use-singular-form-for-collection-repo-folder-name, https://rakyll.org/style-packages/ (section: No plurals), https://github.com/uber-go/guide/blob/master/style.md#package-names

2. "it is common for staging to be structured in the same way as GOPATH", why it is common? do you have any examples?

3. ok, so I'll call it "github.com/clusterpedia.io/api"

27149chen:
@Iceber  thank you for your comments, here is my reply:

1. The reason why I name it api instead of apis is because it is a best practice in Go. See here: https://dmitri.shuralyov.com/idiomatic-go#use-singular-form-for-collection-repo-folder-name, https://rakyll.org/style-packages/ (section: No plurals), https://github.com/uber-go/guide/blob/master/style.md#package-names

2. "it is common for staging to be structured in the same way as GOPATH", why it is common? do you have any examples?

3. ok, so I'll call it "github.com/clusterpedia.io/api"

Iceber:
> The reason why I name it api instead of apis is because it is a best practice in Go. See here: https://dmitri.shuralyov.com/idiomatic-go#use-singular-form-for-collection-repo-folder-name, https://rakyll.org/style-packages/ (section: No plurals), https://github.com/uber-go/guide/blob/master/style.md#package-names

That's right, these are very meaningful references.

> "it is common for staging to be structured in the same way as GOPATH", why it is common? do you have any examples?

Example [kubernetes](https://github.com/kubernetes/kubernetes/tree/master/staging) and [kubeedge](https://github.com/kubeedge/kubeedge/tree/master/staging/src).
An additional benefit is that when generating client code(clientset, informer, lister), we can specify `--output-base` as *. /staging/src*.

hack/update-codegen.sh
```diff
@@ -16,49 +16,55 @@ cd "${REPO_ROOT}"
...
 deepcopy-gen \
-    --go-header-file hack/boilerplate.go.txt \
-    --input-dirs=./pkg/apis/clusterpedia/v1beta1 \
+    --go-header-file ../../../../../hack/boilerplate.go.txt \
+    --input-dirs=./clusterpedia/v1beta1 \
     --output-file-base=zz_generated.deepcopy
 deepcopy-gen \
-    --go-header-file hack/boilerplate.go.txt \
-    --input-dirs=./pkg/apis/clusterpedia \
+    --go-header-file ../../../../../hack/boilerplate.go.txt \
```
Iceber, line 27: It is not recommended to use too many b paths in the script, the whole script is based on REPO_ROOT, it might be clearer to assign the path of the api repo to the variable and then use it in the script.
example: STAGING_REPO_PATH=${REPO_ROOT}/staging/src/github.com/clusterpedia-io

Also based on the repo directory, the following command `cd ../../../../..` can be avoided.

hack/update-crds.sh
```diff
@@ -5,5 +5,5 @@ set -o nounset
 set -o pipefail
 
 echo "Generating CRDs With controller-gen"
-GO11MODULE=on go install sigs.k8s.io/controller-tools/cmd/controller-gen
-controller-gen crd paths=./pkg/apis/cluster/... output:crd:dir=./deploy/crds
+GO11MODULE=on go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0
+cd staging/src/github.com/clusterpedia-io/api && controller-gen crd paths=./cluster/... output:crd:dir=../../../../../deploy/crds
```
Iceber, line 9: Also use `REPO_ROOT` and `STAGING_REPO_ROOT` to avoid `cd taging/src/github.com/clusterpedia-io/api` and `..`

Iceber (commented):
Everything else looks good, Thanks!

hack/update-codegen.sh
27149chen: done

hack/update-crds.sh
27149chen: done

Iceber:
@27149chen Thank you very much for contributing.
We can now think about how to sync api to [api](https://github.com/clusterpedia-io/api)

PR #123, opened 2022-04-01
Supports remaincount as a negative number

cleverhu:
In use, if the offset used is too large, there is no data in the query. The returned remain count is 0, which will make the front-end paging unusable.Supporting remain count as a negative number for paging.

### DIFF
--- a/pkg/storage/internalstorage/resource_storage.go
+++ b/pkg/storage/internalstorage/resource_storage.go
@@ -182,10 +182,9 @@ func (s *ResourceStorage) List(ctx context.Context, listObject runtime.Object, o
 	}
 
 	if amount != nil {
+		// When offset is too large, the data in the response is empty and the remaining count is negative.
+		// This ensures that `amount = offset + len(objects) + remain`
 		remain := *amount - offset - int64(len(objects))
-		if remain < 0 {
-			remain = 0
-		}
 		list.SetRemainingItemCount(&remain)
 	}
 


pkg/storage/internalstorage/resource_storage.go
```diff
@@ -183,9 +183,6 @@ func (s *ResourceStorage) List(ctx context.Context, listObject runtime.Object, o
 
 	if amount != nil {
 		remain := *amount - offset - int64(len(objects))
-		if remain < 0 {
```
kerthcet, line 186: remain can still be 0, right? e.g. `*amount` equals `offset+int64(len(objects))`.

cleverhu: Yes, only when the current page has no data and the offset is too large, a negative number will be returned for front-end paging.

kerthcet: nit: add a comment explaining `remain` might be negative and the reason why this is acceptable.(I think we used to add this logic for specific intension)

Iceber: > nit: add a comment explaining remain might be negative and the reason why this is acceptable.(I think we used to add this logic for specific intension)

That's right.
A negative remaining count is unconventional and some comments need to be added to clarify.

Iceber:
The failed golangci-lint is not related to this pr and has been fixed by https://github.com/clusterpedia-io/clusterpedia/pull/124

PR #126, opened 2022-04-07
update generated crds path

Iceber:
Change the crd yaml generation path to ./charts/_crds


Iceber:
@calvin0327 @wzshiming PTAL

PR #135, opened 2022-04-12
Add test framework

wzshiming:


.github/workflows/test.yml
```diff
@@ -0,0 +1,23 @@
+name: Test
```
Iceber, line 1: `E2E Test` is better? 
Or put it under CI Workflow? After all, the cost of E2E is high, and e2e testing can only be done if the other tests pass

wzshiming: Done, PTAL again

Iceber (approved):
LGTM，Thanks!

Issue #121, opened 2022-03-30
helm: If the release name contains `mysql` or `postgresql` , the clusterpedia will not start successfully


Iceber:
### What happened?

If the release name contains the name of a storage component, then clusterpedia will not start successfully.

realease name: clusterpedia-with-postgresql
```bash
$ helm install clusterpedia-with-postgresql  -n clusterpedia-system --create-namespace  --set persistenceMatchNode=caiwei-master --set installCRDs=true .
```

```bash
╭─icebergu@IceberGudeMBP ~/clusterpedia/charts  ‹add-helm*›
╰─➤  kubectl -n clusterpedia-system get pods
NAME                                                              READY   STATUS             RESTARTS      AGE
clusterpedia-with-postgresql-0                                    1/1     Running            0             13s
clusterpedia-with-postgresql-apiserver-b96d4c599-4tlcs            0/1     CrashLoopBackOff   1 (11s ago)   13s
clusterpedia-with-postgresql-clustersynchro-manager-7f76cd2sgch   0/1     CrashLoopBackOff   1 (11s ago)   13s

╭─icebergu@IceberGudeMBP ~/clusterpedia/charts  ‹add-helm*›
╰─➤  kubectl -n clusterpedia-system logs clusterpedia-with-postgresql-apiserver-b96d4c599-4tlcs  
Failed to init storage: failed to connect to `host=clusterpedia-with-postgresql-postgresql user=postgres database=clusterpedia`: hostname resolving error (lookup clusterpedia-with-postgresql-postgresql on 10.96.0.10:53: no such host)
```

Found that the svc name provided by postgres does not match the host in internalstorage config:
```bash
╭─icebergu@IceberGudeMBP ~/clusterpedia/charts  ‹add-helm*›
╰─➤  kubectl -n clusterpedia-system get svc
NAME                                     TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)    AGE
clusterpedia-with-postgresql             ClusterIP   10.108.157.173   <none>        5432/TCP   4m29s
clusterpedia-with-postgresql-apiserver   ClusterIP   10.99.111.2      <none>        443/TCP    4m29s
clusterpedia-with-postgresql-hl          ClusterIP   None             <none>        5432/TCP   4m29s

╭─icebergu@IceberGudeMBP ~/clusterpedia/charts  ‹add-helm*›
[...]

PR #138, opened 2022-04-15
fix database service name in configmap different with the actual servicename 

ycsk02:
fix database service name in configmap different with the actual servicename when the release name contains mysql or postgresql.

The dependency bitnami chart of postgresql/mysql will use the release name directly when the release name contains the name, for example postgresql or mysql.

The database host name in clusterpedia configmap should be consistent with the servicename generated by bitnami chart.

issue #121

### DIFF
--- a/charts/templates/_helpers.tpl
+++ b/charts/templates/_helpers.tpl
@@ -129,7 +129,16 @@ Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 */}}
 {{- define "clusterpedia.postgresql.fullname" -}}
-{{- printf "%s-%s" .Release.Name "postgresql" | trunc 63 | trimSuffix "-" -}}
+{{- if .Values.postgresql.fullnameOverride -}}
+     {{- .Values.postgresql.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+     {{- $name := default "postgresql" .Values.postgresql.nameOverride -}}
+     {{- if contains $name .Release.Name -}}
+          {{- .Release.Name | trunc 63 | trimSuffix "-" -}}
+     {{- else -}}
+          {{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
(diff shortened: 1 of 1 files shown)


charts/templates/_helpers.tpl
```diff
@@ -129,16 +129,24 @@ Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 */}}
 {{- define "clusterpedia.postgresql.fullname" -}}
+{{- if contains "postgresql" .Release.Name -}}
+{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
```
Iceber, line 133: This does fix https://github.com/clusterpedia-io/clusterpedia/issues/121.

Thinking further, the problem may still occur when `--set postgresql.fullnameOverride=<some>` is set, so we can refer to the [fullname template](https://github.com/bitnami/charts/blob/6a0f3ad9153a79d7f6e66d50494d741d870b9383/bitnami/common/templates/_names.tpl#L21) used by bitnami/postgresql.

Suggestion：
```
{{- define "clusterpedia.postgresql.fullname" -}}
{{- if .Values.postgresql.fullnameOverride -}}
{{- .Values.postgresql.fullnameOverride | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- $name := default "postgresql" .Values.postgresql.nameOverride -}}
{{- if contains $name .Release.Name -}}
{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
{{- end -}}
{{- end -}}
{{- end -}}
```

Iceber (commented):
@ycsk02 Thank you very much for the first contribution!🎉

Good fix for https://github.com/clusterpedia-io/clusterpedia/issues/121, but we can make it even better

charts/templates/_helpers.tpl
ycsk02: Sure, thanks for your suggestion.

Iceber (approved):
LGTM

PR #142, opened 2022-04-18
add ClusterSynchroInitialized status condition

ycsk02:
The plan to fix this issue:
1, Add one type condition: ClusterSynchro. The ClusterSynchro condition has two status: True/False and three reason: InvalidConfig/InitialFailed/Running corresponding to buildClusterConfig/New/Run progress.
2, If cluster config changed, shutdown the synchro, change the Ready contion to Unkown status with reason ClusterSynchroStop, and set the cluster synchro to nil in the manager's synchros map. 
3, If the new progress failed, change the ClusterSynchro condition to False status with reason InitialFailed, with message of the error.
4, After call the synchro run, change the ClusterSynchro condition to True status with reason Running.

#8

### DIFF
--- a/pkg/synchromanager/clustersynchro_manager.go
+++ b/pkg/synchromanager/clustersynchro_manager.go
@@ -9,17 +9,20 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	clusterv1alpha2 "github.com/clusterpedia-io/api/cluster/v1alpha2"
+
 	crdclientset "github.com/clusterpedia-io/clusterpedia/pkg/generated/clientset/versioned"
 	"github.com/clusterpedia-io/clusterpedia/pkg/generated/informers/externalversions"
 	clusterlister "github.com/clusterpedia-io/clusterpedia/pkg/generated/listers/cluster/v1alpha2"
@@ -28,6 +31,8 @@ import (
 )
 
 const ClusterSynchroControllerFinalizer = "clusterpedia.io/cluster-synchro-controller"
+
+const maxClusterSynchroRetry = 5
 
 type Manager struct {
 	runLock sync.Mutex
@@ -182,8 +187,11 @@ func (manager *Manager) processNextCluster() (continued bool) {
 	cluster = cluster.DeepCopy()
 	if err := manager.reconcileCluster(cluster); err != nil {
 		klog.ErrorS(err, "Failed to reconcile cluster", "cluster", name, "num requeues", manager.queue.NumRequeues(key))
-		manager.queue.AddRateLimited(key)
-		return
+		if manager.queue.NumRequeues(key) < maxClusterSynchroRetry {
+			manager.queue.AddRateLimited(key)
+			return
+		}
+		klog.V(2).Infof("Dropping cluster %q out of the queue: %v", key, err)
 	}
 	manager.queue.Forget(key)
 	return
@@ -223,8 +231,8 @@ func (manager *Manager) reconcileCluster(cluster *clusterv1alpha2.PediaCluster) 
 
 	config, err := buildClusterConfig(cluster)
 	if err != nil {
-		// TODO(iceber): update cluster status
 		klog.ErrorS(err, "Failed to build cluster config", "cluster", cluster.Name)
+		manager.UpdateClusterSynchroCondition(cluster.Name, clusterv1alpha2.InvalidConfigConditionReason, "invalid cluster config: "+err.Error(), metav1.ConditionFalse)
 		return nil
 	}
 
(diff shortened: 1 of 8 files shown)


pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -283,25 +307,54 @@ func (manager *Manager) cleanCluster(name string) error {
...
+		}
+		for _, condition := range status.Conditions {
+			meta.SetStatusCondition(&currentClusterStatus.Conditions, condition)
+		}
+
+		cluster.Status = currentClusterStatus
+
+		_, err = manager.clusterpediaclient.ClusterV1alpha2().PediaClusters().UpdateStatus(ctx, cluster, metav1.UpdateOptions{})
+		if err == nil {
+			klog.V(2).InfoS("Update Cluster Status", "cluster", cluster.Name, "status", status.Conditions[0].Reason)
```
Iceber, line 353: Does this log information need to be modified, the valid information does not feel enough?

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -7,6 +7,10 @@ import (
 	"sync"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/meta"
+
+	"k8s.io/client-go/util/retry"
+
```
Iceber, line 13: Looks like a redundant line break

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -234,29 +244,43 @@ func (manager *Manager) reconcileCluster(cluster *clusterv1alpha2.PediaCluster)
...
+			if err == nil {
+				return true, nil
+			}
+
+			manager.UpdateClusterSynchroCondition(cluster.Name, clusterv1alpha2.InitialFailedConditionReason, err.Error(), metav1.ConditionFalse)
 
-			klog.ErrorS(err, "Failed to create cluster synchro", "cluster", cluster.Name)
+			currentCluster, err := manager.clusterlister.Get(cluster.Name)
+			if err != nil || currentCluster.Generation != cluster.Generation {
+				return false, errors.New("cluster config is changed or failed to get cluster")
```
Iceber, line 269: If the cluster changes, the cluster name already exists in the queue waiting for the next transfer and does not need to be re-queued

Iceber (commented):
Very useful feature.

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -223,8 +233,8 @@ func (manager *Manager) reconcileCluster(cluster *clusterv1alpha2.PediaCluster)
 
 	config, err := buildClusterConfig(cluster)
 	if err != nil {
-		// TODO(iceber): update cluster status
 		klog.ErrorS(err, "Failed to build cluster config", "cluster", cluster.Name)
+		manager.UpdateClusterSynchroCondition(cluster.Name, clusterv1alpha2.InvalidConfigConditionReason, "invalid cluster config: "+err.Error(), metav1.ConditionFalse)
```
Iceber, line 237: If the cluster configuration is modified, but the configuration build fails, do we need to stop synchro？
I'm not sure

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -234,29 +244,43 @@ func (manager *Manager) reconcileCluster(cluster *clusterv1alpha2.PediaCluster)
...
 
 	// create resource synchro
 	if synchro == nil {
-		// TODO(iceber): set the stop sign of the manager to cluster synchro
-		synchro, err = clustersynchro.New(cluster.Name, config, manager.storage, manager)
-		if err != nil {
-			// TODO(iceber): update cluster status
-			// There are many reasons why creating a cluster synchro can fail.
-			// How do you gracefully handle different errors?
+		if err = wait.PollImmediate(waitClusterSynchroInternal, waitClusterSynchroTimeout, func() (done bool, err error) {
```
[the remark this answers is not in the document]
ycsk02: changed to use requeuing for failed reconciling.

pkg/synchromanager/clustersynchro_manager.go
ycsk02: Log the whole status info: type, statue, reason, message.

pkg/synchromanager/clustersynchro_manager.go
ycsk02: The error is not returned for controller requeue. When an error message is received, wait.PollImmediate will stop try and return err,  the err handler will log it and return nill to finish the reconcile progress.

pkg/synchromanager/clustersynchro_manager.go
ycsk02: When a user uses an invalid configuration, we assume that the user's changes have not been completed, so the invalid configuration should not affect the system.

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -283,25 +293,54 @@ func (manager *Manager) cleanCluster(name string) error {
...
+			currentClusterStatus.Version = status.Version
+		}
+		if status.SyncResources != nil {
+			currentClusterStatus.SyncResources = status.SyncResources
+		}
+		for _, condition := range status.Conditions {
+			meta.SetStatusCondition(&currentClusterStatus.Conditions, condition)
+		}
+
+		cluster.Status = currentClusterStatus
```
Iceber, line 335: It is understandable to use `currentClusterStatus` to indicate the current status.
But using `cluster.Status` feels more concise and does not cause other misunderstandings
eg.
```go

                cluster = cluster.DeepCopy()
                if status.Version != "" {
                        cluster.Status.Version = status.Version
                }
                if status.SyncResources != nil {
                        cluster.Status.SyncResources = status.SyncResources
                }
                for _, condition := range status.Conditions {
                        meta.SetStatusCondition(&cluster.Status.Conditions, condition)
                }
```

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -283,25 +293,54 @@ func (manager *Manager) cleanCluster(name string) error {
...
+
+		_, err = manager.clusterpediaclient.ClusterV1alpha2().PediaClusters().UpdateStatus(ctx, cluster, metav1.UpdateOptions{})
+		if err == nil {
+			klog.V(2).InfoS("Update Cluster Status", "cluster", cluster.Name, "type", status.Conditions[0].Type, "status", status.Conditions[0].Status, "reason", status.Conditions[0].Reason, "message", status.Conditions[0].Message)
```
Iceber, line 339: It may be necessary to show the ReadyCondition information as well.
I think it would be better for the caller of  `UpdateClusterStatus` to print the log, since they know which state may have been modified.

This can be fixed separately in other pr.

pkg/synchromanager/clustersynchro_manager.go
ycsk02: ok

pkg/synchromanager/clustersynchro_manager.go
ycsk02: ok

Iceber (approved):
LGTM，Thank you very much for contributing!

PR #158, opened 2022-04-24
Fix frequent cluster status updates

ycsk02:
The cluster status is updated only if the updated state is different from the current state.

### DIFF
--- a/pkg/synchromanager/clustersynchro_manager.go
+++ b/pkg/synchromanager/clustersynchro_manager.go
@@ -315,10 +315,7 @@ func (manager *Manager) UpdateClusterStatus(ctx context.Context, name string, st
 			return err
 		}
 
-		if equality.Semantic.DeepEqual(&cluster.Status, status) {
-			return nil
-		}
-
+		lastStatus := cluster.Status
 		cluster = cluster.DeepCopy()
 
 		if status.Version != "" {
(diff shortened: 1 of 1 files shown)


ycsk02:
@Iceber PTAL, thanks!

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -315,22 +315,24 @@ func (manager *Manager) UpdateClusterStatus(ctx context.Context, name string, st
...
 		if status.SyncResources != nil {
-			cluster.Status.SyncResources = status.SyncResources
+			expectStatus.SyncResources = status.SyncResources
 		}
 		for _, condition := range status.Conditions {
-			meta.SetStatusCondition(&cluster.Status.Conditions, condition)
+			meta.SetStatusCondition(&expectStatus.Conditions, condition)
+		}
+
+		cluster = cluster.DeepCopy()
```
Iceber, line 329: It looks like it only needs to be deepcopied once.
eg.
```go

                lastStatus := cluster.Status
                cluster = cluster.DeepCopy()

                if status.Version != "" {
                        cluster.Status.Version = status.Version
                }
                if status.SyncResources != nil {
                        cluster.Status.SyncResources = status.SyncResources
                }
                for _, condition := range status.Conditions {
                        meta.SetStatusCondition(&cluster.Status.Conditions, condition)
                }

                if equality.Semantic.DeepEqual(cluster.Status, lastStatus) {
                        return nil
                }
```
What do you think?

ycsk02: Exactly

Iceber (approved):
LGTM

PR #160, opened 2022-04-25
support raw sql query

cleverhu:
Support raw sql by query path and add feature-gate for it。

How to use:
First: Add ` --feature-gates=AllowRawSQLQuery=true` args for apiserver deployment to open the feature-gate.

Second:
Use kubectl:
`kubectl get --raw="/apis/clusterpedia.io/v1beta1/resources/apis/apps/v1/namespaces/default/deployments?limit=5&whereSQL=(namespace='default') or (name like '%25b%25')"`
tips:
%25 is the % after encode
or use clusterpedia client-go.
image
image

Known problems:
1、Preventing SQL injection is needed to be done.
2、If and incorrect SQL statement is passed in. If it is a number, ID will be used as the query condition by default.
If it contains English letter, it will be passed in as column.
such as:
A number is passed in:
wecom-temp-e15867c7cedc1bca0a5fde21dc48d226
English letter is passed in:
wecom-temp-8509c799f282d1b2ca439ea0f794d9bb

### DIFF
--- a/pkg/storage/internalstorage/features.go
+++ b/pkg/storage/internalstorage/features.go
@@ -1,0 +1,25 @@
+package internalstorage
+
+import (
+	"k8s.io/apimachinery/pkg/util/runtime"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
+	"k8s.io/component-base/featuregate"
+)
+
+const (
+	// AllowRawSQLQuery is a feature gate for the apiserver to allow querying by the raw sql.
+	//
+	// owner: @cleverhu
+	// alpha: v0.3.0
+	AllowRawSQLQuery featuregate.Feature = "AllowRawSQLQuery"
+)
+
+func init() {
+	runtime.Must(utilfeature.DefaultMutableFeatureGate.Add(defaultInternalStorageFeatureGates))
+}
+
+// defaultInternalStorageFeatureGates consists of all known custom internalstorage feature keys.
+// To add a new feature, define a key for it above and add it here.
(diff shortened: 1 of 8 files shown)


Iceber:
LGTM

> If and incorrect SQL statement is passed in. If it is a number, ID will be used as the query condition by default.
If it contains English letter, it will be passed in as column.

Could you give some examples.

pkg/storage/internalstorage/features.go
```diff
@@ -0,0 +1,25 @@
...
+	AllowRawSQLQuery featuregate.Feature = "AllowRawSQLQuery"
+)
+
+func init() {
+	runtime.Must(utilfeature.DefaultMutableFeatureGate.Add(defaultInternalstorageFeatureGates))
+}
+
+// defaultInternalstorageFeatureGates consists of all known custom internalstorage feature keys.
+// To add a new feature, define a key for it above and add it here.
+var defaultInternalstorageFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
```
Iceber, line 23: typo: `defaultInternalstorageFeatureGates` -> `defaultInsternalStorageFeatureGates`

calvin0327:
Should we put  featureGate file all together in one file？we can submit another PR if needed.

Iceber:
> Should we put featureGate file all together in one file？we can submit another PR if needed.

This feature is specific to the storage layer plugin internalstorage and should not be mixed with generic packages like clustersynchro and apiserver

Issue #6, opened 2021-12-02
support discovery of abnormal states during resource synchronization


Iceber:
During resource synchronization, the `informer reflector` cannot report errors externally if a resource(custom resource) is deleted, or after not having the correct resource permissions.

We need to make the `reflector` able to tell the `resource synchro` more information

ycsk02:
Abnormal resource synchronization information can be displayed in SyncResources of cluster status
SharedIndexInformer supports setting a custom error handling function by the Config item WatchErrorHandler.  
Add WatchErrorHandler in the resourceVersionInformer to specify the custom error handler. Due to reflector's watchErrorHandler is a unexported field, it requires a custom reflector.  


```
func ErrorHandlerForResourceSynchro(synchro *ResourceSynchro) informer.WatchErrorHandler {
	return func(_ *informer.Reflector, err error) {
		status := clusterv1alpha2.ClusterResourceSyncCondition{
			Status:             clusterv1alpha2.SyncStatusError,
			Reason:             err.Error(),
			LastTransitionTime: metav1.Now().Rfc3339Copy(),
		}
		synchro.status.Store(status)
	}
}

	informer.NewResourceVersionInformer(
		synchro.cluster,
		synchro.listerWatcher,
		synchro.cache,
		synchro.example,
		synchro,
	).SetWatchErrorHandler(ErrorHandlerForResourceSynchro(synchro)).Run(informerStopCh)

	config := &Config{
		ListerWatcher: lw,
		ObjectType:    exampleObject,
		RetryOnError:  false,
		Process: func(obj interface{}) error {
			deltas := obj.(cache.Deltas)
			return informer.HandleDeltas(deltas)
		},
		Queue: cache.NewDeltaFIFOWithOptions(cache.DeltaFIFOOptions{
			KeyFunction:           cache.DeletionHandlingMetaNamespaceKeyFunc,
			KnownObjects:          informer.storage,
			EmitDeltaTypeReplaced: true,
		}),
		WatchErrorHandler: informer.watchErrorHandler,
	}
```

```
  Sync Resources:
    Group:  apps
    Resources:
      deployments
    Group:
    Resources:
      pods
    Group:  iam.aiscope
    Resources:
      users
    Versions:
      v1alpha2
Status:
  Conditions:
    Last Transition Time:  2022-04-25T07:50:17Z
    Message:
    Reason:                Running
    Status:                True
    Type:                  ClusterSynchroInitialized
[...]

PR #161, opened 2022-04-26
support discovery of abnormal states during resource synchronization

ycsk02:
1, retrive resourceNegotiator.restmapper in setSyncResources
2, don't delete the version in groupResourceStatus when it is in cluster syncResources
3, add watchErrorHandler ErrorHandlerForResourceSynchro to customize reflector
4, add SyncStatusError status to ClusterResourceSyncCondition

#6

### DIFF
--- a/pkg/synchromanager/clustersynchro/resource_synchro.go
+++ b/pkg/synchromanager/clustersynchro/resource_synchro.go
@@ -198,21 +198,16 @@ func (synchro *ResourceSynchro) Start(stopCh <-chan struct{}) {
 		close(informerStopCh)
 	}()
 
-	status := clusterv1alpha2.ClusterResourceSyncCondition{
-		Status:             clusterv1alpha2.SyncStatusSyncing,
-		LastTransitionTime: metav1.Now().Rfc3339Copy(),
-	}
-	synchro.status.Store(status)
-
 	informer.NewResourceVersionInformer(
 		synchro.cluster,
 		synchro.listerWatcher,
 		synchro.cache,
 		synchro.example,
 		synchro,
+		ErrorHandlerForResourceSynchro(synchro),
 	).Run(informerStopCh)
 
-	status = clusterv1alpha2.ClusterResourceSyncCondition{
+	status := clusterv1alpha2.ClusterResourceSyncCondition{
 		Status:             clusterv1alpha2.SyncStatusStop,
 		Reason:             "Pause",
 		LastTransitionTime: metav1.Now().Rfc3339Copy(),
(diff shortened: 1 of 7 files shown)


pkg/synchromanager/clustersynchro/resource_negotiator.go
```diff
@@ -103,6 +103,10 @@ func (negotiator *ResourceNegotiator) negotiatorVersions(resource schema.GroupRe
 	isLegacyResource := resourcescheme.LegacyResourceScheme.IsGroupRegistered(resource.Group)
 	if !isLegacyResource {
 		mapper, supportedVersions = negotiator.customResourceController.GetRESTMappingAndVersions(resource)
+		if mapper == nil {
+			klog.InfoS("Skip resource sync", "cluster", negotiator.name, "resource", resource, "reason", "No matches for resource")
+			return nil, nil, isLegacyResource
+		}
```
Iceber, line 109: `LegacyResourceScheme` represents the resources supported by the current kubernetes code dependency.
If `isLegacyResource` equals false, it does not necessarily mean that it is a custom resources, or it could be a new resource added to the cluster that is higher than the dependency version.

For such a resource, we need to go to the restmapper to get it, but isLegacyResource is false.

Here the code omits to determine whether it is a custom resource, but to simplify the code by whether it is possible to get mapping about the custom resource, and does not add comments, which may lead to some misunderstandings

ycsk02: Ok, got it.

ycsk02:
when the status is not changed, genClusterStatus will update resource SyncCondition LastTransitionTime in order not to update frequently.

Unknown resources cannot distinguish between custom resources and new resources for the dependency version. A new PR will be created for detailed description.

@Iceber PTAL, Thanks.

pkg/synchromanager/clustersynchro/cluster_synchro.go
```diff
@@ -322,6 +323,15 @@ func (s *ClusterSynchro) setSyncResources() {
...
 		storageGVR := key.(schema.GroupVersionResource)
+		for _, groupResources := range syncResources {
+			if hasString(groupResources.Resources, storageGVR.Resource) {
+				if groupResources.Versions != nil && !hasString(groupResources.Versions, storageGVR.Version) {
+					return true
+				}
+				delete(deleted, storageGVR)
+				return true
+			}
+		}
```
Iceber, line 334: It does not seem to have much to do with this pr, you can create a new pr and describe it.
This allows the pr to stay focused, making it easier to review and merge.

ycsk02: delete(deleted, storageGVR) will keep the ResourceSynchro in storageResourceSynchros. genClusterStatus generates the ResourceSynchro status by storageResourceSynchros.
if we removed ResourceSynchro in storageResourceSynchros, the error message reported by reflector cann't be displayed.

ycsk02:
> when the status is not changed, genClusterStatus will update resource SyncCondition LastTransitionTime in order not to update frequently.
> 

Based on discussions with @Iceber , it is not recommended to do save operation groupResourceStatus.Store outside of setSyncResources, the resource SyncCondition status should dynamic compute from synchro.

> In setSyncResources, will keep the resource in storageResourceSynchros even though it is not in storageResourceSyncConfigs. The resource Informer will keep running.
> 

When user delete the custom resource in Kubernetes, we just remove the resource status from cluster, it does not need to do any process.  We are more concerned with scenarios without permissions in issue #6 

> In Reflector.Run, will always call watchErrorHandler, when starting and an error occurs.
> 

It is not optimal and a better approach will be considered later.

pkg/synchromanager/clustersynchro/resource_synchro.go
```diff
@@ -400,3 +395,22 @@ func (synchro *ResourceSynchro) deleteResource(obj runtime.Object) error {
...
+		}
+		if err != nil {
+			status = clusterv1alpha2.ClusterResourceSyncCondition{
+				Status:             clusterv1alpha2.SyncStatusError,
+				Reason:             "ResourceWatchFailed",
+				Message:            err.Error(),
+				LastTransitionTime: metav1.Now().Rfc3339Copy(),
+			}
+		}
+		synchro.status.Store(status)
```
Iceber, line 414: Do we need to call the `DefaultWatchErrorHandler ` again so that it may be more beneficial when analyzing the logs separately?

ycsk02: Yes, good idea.

Iceber:
This pr has been resolved https://github.com/clusterpedia-io/clusterpedia/issues/6, but we still need to optimize the WatchErrorHandler in the reflector.
You can also copy the original `reflector.go` to *informer/.reflector.go* for others to compare the changes.

Iceber:
@ycsk02 Thank you very much for contributing!

Issue #81, opened 2022-02-15
Use feature gates to control whether the count of remaining items is allowed to be returned


Iceber:
Currently, if  need to return the `RemainingItemCount`, we need to set `withRemaingCount` in the `Search Label` or `url query` https://github.com/clusterpedia-io/clusterpedia/pull/33

It is now possible to allow all requests to return the `RemainingItemCount` by reusing [the feature gate of the kube apiserver -- `RemainingItemCount`](https://github.com/kubernetes/kubernetes/blob/d899c39ca3025362033b5a71e4f27c32690dc78b/staging/src/k8s.io/apiserver/pkg/features/kube_features.go#L94)

calvin0327:
I think it's better if the both of those way is be supported.  the `RemainingItemCount` is not required very much. what do you think?

Iceber:
@calvin0327 Yes. Use feature gate to control the default behaviour and also support active setting in the `url query` or `search label`

Iceber:
@feeltimeQ Thank you for joining!

Iceber:
The default value for feature gateway RemainingItemCount is true.

We can't change this default, so it seems we need to actively set it in the charts and  apiserver deploy yaml. `--feature-gates=RemainingItemCount=false`.
https://github.com/clusterpedia-io/clusterpedia/blob/e38f39f03d2483ebbfc5b2c96a13e6b71a0f835f/deploy/clusterpedia_apiserver_deployment.yaml#L36-L44
What do you think? @calvin0327 @cleverhu

PR #166, opened 2022-04-27
feature/use RemainingItemCount feature gate

feeltimeQ:
## issue: #81

if RemainingItemCount=true
wecom-temp-b80f1520bbcabf1829850fe674995cb8

if RemainingItemCount=false
wecom-temp-d653afaa4db30d7ddde498ad59d7f71a

### DIFF
--- a/pkg/kubeapiserver/resourcerest/storage.go
+++ b/pkg/kubeapiserver/resourcerest/storage.go
@@ -10,8 +10,10 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	genericrequest "k8s.io/apiserver/pkg/endpoints/request"
+	genericfeatures "k8s.io/apiserver/pkg/features"
 	"k8s.io/apiserver/pkg/registry/rest"
 	storeerr "k8s.io/apiserver/pkg/storage/errors"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
 
 	internal "github.com/clusterpedia-io/api/clusterpedia"
 	"github.com/clusterpedia-io/api/clusterpedia/scheme"
@@ -87,11 +89,11 @@ func (s *RESTStorage) list(ctx context.Context, options *internal.ListOptions) (
 		return nil, apierrors.NewBadRequest("If searching by owner uid or name, then the cluster must be specified")
 	}
 
-	/*
-		TODO(iceber): add feature gates to set default option
-		if options.WithRemainingCount == nil {
+	if options.WithRemainingCount == nil {
+		if enabled := utilfeature.DefaultFeatureGate.Enabled(genericfeatures.RemainingItemCount); enabled {
+			options.WithRemainingCount = &enabled
 		}
-	*/
+	}
 
 	objs := s.NewList()
 	if err := s.Storage.List(ctx, objs, options); err != nil {


cleverhu (commented):
When the user set the WithRemainingCount to false, and does not want to return the remaining count in order to speed up the search efficiency, it will waste redundant resources. Therefore, it is necessary to judge whether the WithRemainingCount is nil.

pkg/kubeapiserver/resourcerest/storage.go
```diff
@@ -70,6 +72,7 @@ func (s *RESTStorage) List(ctx context.Context, _ *metainternalversion.ListOptio
 }
 
 func (s *RESTStorage) list(ctx context.Context, options *internal.ListOptions) (runtime.Object, error) {
+	defaultRemainingItemCount := utilfeature.DefaultFeatureGate.Enabled(defaultfeatures.RemainingItemCount)
```
cleverhu, line 75: This value is required when remaincount is nil, you don't need to judge at the beginning every time.

cleverhu: The variable defaultremainingitemcount means whether it is enabled. Using this name will be a little tempting.

cleverhu:
It is suggested that query should also be supported. There is still some extra work to be done.

pkg/kubeapiserver/resourcerest/storage.go
```diff
@@ -10,8 +10,10 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	genericrequest "k8s.io/apiserver/pkg/endpoints/request"
+	defaultfeatures "k8s.io/apiserver/pkg/features"
```
Iceber, line 13: The usual alias is `genericfeatures`

pkg/kubeapiserver/resourcerest/storage.go
```diff
@@ -87,11 +89,12 @@ func (s *RESTStorage) list(ctx context.Context, options *internal.ListOptions) (
 		return nil, apierrors.NewBadRequest("If searching by owner uid or name, then the cluster must be specified")
 	}
 
-	/*
-		TODO(iceber): add feature gates to set default option
-		if options.WithRemainingCount == nil {
+	if options.WithRemainingCount == nil {
+		if utilfeature.DefaultFeatureGate.Enabled(defaultfeatures.RemainingItemCount) != false {
```
Iceber, line 93: Suggestion:
```go
if enabled := utilfeature.DefaultFeatureGate.Enabled(defaultfeatures.RemainingItemCount); enabled {
    options.WithRemainingCount = &enabled
}
```

Iceber (commented):
Commit looks like a branch name that can be changed to `feature: use RemainingItemCount feature gate`

DCO Check is failed.
> 
> 
> Here is how to fix the problem so that this code can be merged.

Iceber (approved):
LGTM

Iceber:
@calvin0327 @cleverhu The default value for feature gateway `RemainingItemCount` is true.
We can't change this default, so it seems we need to actively set these in the charts and deploy yaml. `--feature-gates=RemainingItemCount=true

Iceber:
The default value for feature gateway RemainingItemCount is true.

We can't change this default, so it seems we need to actively set it in the charts and  apiserver deploy yaml. `--feature-gates=RemainingItemCount=false`.
https://github.com/clusterpedia-io/clusterpedia/blob/e38f39f03d2483ebbfc5b2c96a13e6b71a0f835f/deploy/clusterpedia_apiserver_deployment.yaml#L36-L44
What do you think? @calvin0327 @cleverhu

PR #170, opened 2022-04-30
fix reload restmapper when crd deleted

ycsk02:
Add restmapper in CustomResourceController, when a CRD resource is been deleted, trigger to reload the cluster restermapper by  querying a never exists resource version. The dynamicRESTMapper method checkAndReload will reload the mapper at the backend.

@Iceber PTAL, thanks.

#6


Iceber:
LGTM, This does ensure that CRDs are not queried from the restmapper again after they have been deleted.

Issue #144, opened 2022-04-19
When sorting the resource_version field in the cluster, the result is wrong


liuzouzou:
### What happened?

<img width="1222" alt="wecom-temp-6481f3f543fe2366435429f32712a5ba" src="https://user-images.githubusercontent.com/43669330/163936539-63c08a8b-8ebd-412f-916d-3a477ddb761a.png">
<img width="1245" alt="wecom-temp-af951f2228022fa10d9553e247e0b89e" src="https://user-images.githubusercontent.com/43669330/163936662-13292127-0122-420e-9c00-ee37e057ef12.png">

### What did you expect to happen?

When sorting the resource_version field in the cluster, the query results are correct

### How can we reproduce it (as minimally and precisely as possible)?

kubectl --cluster clusterpedia get pods -l  "search.clusterpedia.io/orderby in (resource_version_desc)"
kubectl --cluster clusterpedia get pods -l  "search.clusterpedia.io/orderby in (resource_version)"

### Clusterpedia Version

<details>

```console
# replace the pod name of clusterpedia apiserver
$ kubectl -n clusterpedia-system exec {clusterpedia-apiserver-pod-name} -- apiserver --version=raw
# paste output here

# replace the pod name of clusterpedia clustersynchro-manager
$ kubectl -n clusterpedia-system exec {clustersynchro-manager-pod-name} -- clustersynchro-manager --version=raw
# paste output here
```

</details>

### Host Kubernetes version

<details>

```console
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.20", GitCommit:"1f3e19b7beb1cc0110255668c4238ed63dadb7ad", GitTreeState:"clean", BuildDate:"2021-06-16T12:58:51Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.20", GitCommit:"7c28a83e67cb506f9df948d7b23698f8f9c279b8", GitTreeState:"clean", BuildDate:"2021-07-15T08:58:53Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

</details>

Iceber:
@liuzouzou Thank you very much for the feedback!

The `internalstorage` saves the resource version as a string, but does not convert it to a number for sorting.

We need to convert the `resource_version` field to int when sorting.
[...]

PR #171, opened 2022-05-01
Fix orderby resourceVersion

RuliXu:
Fix https://github.com/clusterpedia-io/clusterpedia/issues/144


pkg/storage/internalstorage/util.go
```diff
@@ -135,9 +135,13 @@ func applyListOptionsToQuery(query *gorm.DB, opts *internal.ListOptions, applyFn
 	// Due to performance reasons, the default order by is not set.
 	// https://github.com/clusterpedia-io/clusterpedia/pull/44
 	for _, orderby := range opts.OrderBy {
-		if supportedOrderByFields.Has(orderby.Field) {
+		orderBy := orderby.Field
```
Iceber, line 138: Since there is already a variable `orderby` with a similar name, it might be easier to read by naming it `field` or `column` (`field` maybe better).

Iceber (approved):
LGTM
@RuliXu Thank you very much for contributing :tada:! Expect more participation from you.

RuliXu:
> LGTM @RuliXu Thank you very much for contributing 🎉! Expect more participation from you.

Thank you for your help! 🎉 With pleasure.

PR #172, opened 2022-05-04
[BUG] - External Database Password

gambol99:
Note, I'm not sure if this is intentional so raising as a bug. The external password is not base64 encoded and so is copied direct into the secret. 

On a related issue, since I was considering to raise it - it would be nice to make the name `name` amd `key` configurable as it makes it eaiser to share between components and potentially supporting a DB_HOSTNAME for again easy templating. 

```YAML
- name: DB_PASSWORD
  valueFrom:
     secretKeyRef:
       name: {{ include "clusterpedia.internalstorage.fullname" . }}
       key: password
```

### DIFF
--- a/charts/templates/_helpers.tpl
+++ b/charts/templates/_helpers.tpl
@@ -68,7 +68,7 @@ Return the proper Docker Image Registry Secret Names
 
 {{- define "clusterpedia.storage.password" -}}
 {{- if eq .Values.storageInstallMode "external" }}
-     {{- required "Please set correct storage password!" .Values.externalStorage.password -}}
+     {{- required "Please set correct storage password!" .Values.externalStorage.password | b64enc -}}
 {{- else -}}
      {{- if eq (include "clusterpedia.storage.type" .) "postgres" }}
           {{- if not (empty .Values.global.postgresql.auth.username) -}}


Iceber:
LGTM
@gambol99 Thank you very much for finding and fixing this bug!

Iceber:
> it would be nice to make the name name amd key configurable as it makes it eaiser to share between components

Good idea, this allows we to reuse the same secret without having to create an additional secret.

> potentially supporting a DB_HOSTNAME for again easy templating.

However, host is usually set in internalstorage-config, so if environment variables are supported, how do you reconcile config and environment variables?
Do we not set the Host in the internalstorage config, and then set it in the environment variable?

calvin0327:
LGTM

Iceber:
> > it would be nice to make the name name amd key configurable as it makes it eaiser to share between components
> 
> Good idea, this allows we to reuse the same secret without having to create an additional secret.
> 
> > potentially supporting a DB_HOSTNAME for again easy templating.
> 
> However, host is usually set in internalstorage-config, so if environment variables are supported, how do you reconcile config and environment variables? Do we not set the Host in the internalstorage config, and then set it in the environment variable?

@gambol99 The pr has been merged, we can discuss it in a new issue and you are more than welcome to continue participating!

PR #173, opened 2022-05-05
move fuzzyname search to applyListOptionsToQuery

cleverhu:
move fuzzyname search to applyListOptionsToQuery


Iceber (approved):
LGTM

Fuzzy search is a more general function that can also be used by collection resources, although only search labels are currently supported

Issue #116, opened 2022-03-17
Support find Deploy/StatefulSet/DaemonSet related to pods


RuliXu:
### What would you like to be added?

Support find out which Deploy/StatefulSet/DaemonSet a pod belongs to.

### Why is this needed?

We would like to have a pod list related to a specific service, and the workload(deploy/statefulset/daemonset) related to the pod. Of course we can get the related workloads with the service selector, but we can only get pods and workloads separately, we can't tell directly which workload one pod belongs to.

Iceber:
Currently, it is supported to find related pods through resources such as deployment/daemonset.
Do you want to find the corresponding deployment or daemonset through pod?

It is now possible to use the [field selector](https://clusterpedia.io/docs/usage/search/multi-cluster/#field-selector) to find the corresponding deployment or daemonset through the pod's lables.
`pod.labels -> deployment.spec.selector.matchLabels`

Does using field selector meet your needs?

RuliXu:
> Currently, it is supported to find related pods through resources such as deployment/daemonset. Do you want to find the corresponding deployment or daemonset through pod?

Yes
> It is now possible to use the [field selector](https://clusterpedia.io/docs/usage/search/multi-cluster/#field-selector) to find the corresponding deployment or daemonset through the pod's lables. `pod.labels -> deployment.spec.selector.matchLabels`
> 
> Does using field selector meet your needs?

I‘m afraid not, for example if pod.labels are more than deployment.spec.selector.matchLabels. I'm thinking we need to find corresponding workload through pod's ownerference.

Iceber:
Usually when we have obtained the complete pod or metadata, and then use metadata.owner to determine the workload it belongs to (if it is a ReplicaSet, then we could search Deployment, if it is DaemonSet/Statefulset/Job, then directly search Daemonset/StatefulSet/Job) .

Do you want to directly use the pod name to get the corresponding workload information without getting the complete resources or metadata of the specific pod?

RuliXu:
> Usually we have obtained the complete pod or metadata, and then use metadata.owner to determine the workload it belongs to (if it is a ReplicaSet, then we could search Deployment, if it is DaemonSet/Statefulset/Job, then directly search Daemonset/StatefulSet/Job) .

Yes, like this.
> Do you want to directly use the pod name to get the corresponding workload information without getting the complete resources or metadata of the specific pod?

Not only the pod name, but the full data of Pod could be used.

PR #175, opened 2022-05-06
fix frequently status switch when listwatch error

ycsk02:
When reflector ListAndWatch occurs error, it will repeatedly run ListAndWatch. So the ClusterResourceSyncCondition status will frequently switch from Syncing to Error. 

Change call watchErrorHandler in watchHandler to set status Syncing, after Watch connection is established.

Add original reflector.go.

@Iceber PTAL, thanks.

#116 

#6

### DIFF
--- a/pkg/synchromanager/clustersynchro/informer/.reflector.go
+++ b/pkg/synchromanager/clustersynchro/informer/.reflector.go
@@ -1,4 +1,20 @@
-package informer
+/*
+Copyright 2014 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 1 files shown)


Iceber:
@ycsk02 What do you think about the problem that the informer handler may have been notified of the data after the list before the watch started?
But until there is a better solution, the current fix is fine and better than before ListAndWatch

ycsk02:
> 

Hi @Iceber , we expect to be able to present a state of resource synchronization, such as synchronization `Syncing` or failure `Error`.  List is the process of resource synchronization to obtain the full amount of data for the first time. This process does not represent a change in the state of our resource synchronization, but only indicates that we have received a batch of correct data.  Only after the Watch connection is successfully established, can we indicate that the state of resource synchronization is in synchronization.

Iceber (approved):
LGTM

Issue #167, opened 2022-04-28
upgrade featuregate `PruneLastAppliedConfiguration` `PruneManagedFields` to beta in v0.0.3


calvin0327:
### What would you like to be added?

 In the past few releases, both featrue `PruneLastAppliedConfiguration` and `PruneManagedFields` have been stable and mature. Will we update these featuregate in V0.0.3?

### Why is this needed?

feature upgrade

Iceber:
These two features are available in 0.0.9-alpha, and I feel they are very useful, and can be upgraded to beta in 0.3.0, by default to True.

@ycsk02 @27149chen @wawa0210 What do you think, or will we upgrade to beta at 0.4 or higher?

27149chen:
It's ok to me

wawa0210:
The default featuregate is very useful, greatly reduces storage space, and can be released as soon as possible if possible

ycsk02:
useful featuregate, looks good.

Iceber:
cc @calvin0327 
Please upgrade these feature gates.

PR #176, opened 2022-05-06
update the featrueGate PruneManagedFields and PruneLastAppliedConfiguration to beta

calvin0327:
fixed: https://github.com/clusterpedia-io/clusterpedia/issues/167

update the featrueGate PruneManagedFields and PruneLastAppliedConfiguration to beta

/kind feature
/cc @Iceber

### DIFF
--- a/pkg/synchromanager/features/features.go
+++ b/pkg/synchromanager/features/features.go
@@ -19,12 +19,14 @@ const (
 	//
 	// owner: @iceber
 	// alpha: v0.0.9
+	// beta: v0.3.0
 	PruneManagedFields featuregate.Feature = "PruneManagedFields"
 
 	// PruneLastAppliedConfiguration is a feature gate for the ClusterSynchro to prune `LastAppliedConfiguration` of the resource
 	//
 	// owner: @iceber
 	// alpha: v0.0.9
+	// beta: v0.3.0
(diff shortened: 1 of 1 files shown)


charts/templates/clustersynchro-manager-deployment.yaml
```diff
@@ -37,7 +37,6 @@ spec:
         - /usr/local/bin/clustersynchro-manager
         - --storage-config=/etc/clusterpedia/storage/internalstorage-config.yaml
         - --leader-elect-resource-namespace={{ .Release.Namespace }}
-        - --feature-gates=PruneManagedFields=true,PruneLastAppliedConfiguration=true
```
Iceber, line 40: Would it be better to set it explicitly for the feature gate we define and use?

calvin0327: The feature gate is beta at current stage, we can set it explicitly.  if the feature gate update to GA, we don't have to specify it, do you think?

Iceber: > The feature gate is beta at current stage, we can set it explicitly. if the feature gate update to GA, we don't have to specify it

I think it's fine

Iceber (approved):
LGTM

PR #178, opened 2022-05-07
fix: json tag err

yyzxw:
fix: json tag err


Iceber:
@yyzxw Thank you so much for fixing it!

DCO Check is failed.
> 
> 
> Here is how to fix the problem so that this code can be merged.

Iceber (approved):
LGTM, This field is not currently in use and does not raise compatibility problems

PR #179, opened 2022-05-07
fix the CRD file is missing from the deploy directory

liangyongzhenya:
fix the CRD file is missing from the deploy directory


calvin0327:
@liangyongzhenya we may want to soft link CRD files to this directory

Iceber:
LGTM，Thank you very much for contributing

Issue #168, opened 2022-04-30
Support wildcard (like "*") for resources to sync


27149chen:
### What would you like to be added?

Support wildcard (like "*") for resources to sync, in the following format:
```yaml
# all resources in all groups
syncResources:
- group: "*"
  resources:
  - "*"
  ```

```yaml
# all resources in group apps
syncResources:
- group: "apps"
  resources:
  - "*"
  ```

### Why is this needed?

we need to list all resources in a cluster dynamically (the resource types might be changing)

Iceber:
Wildcard is useful.
For synchronizing all resources in all groups, the feature gate `AllowSyncALLResources`  is added to enable wildcard for groups.

The wildcard for resources in a group are supported by default

Iceber:
In addition, resources that do not support the watch method are still not synchronized

PR #180, opened 2022-05-09
Dynamic discovery of changes in resource types

Iceber:
https://github.com/clusterpedia-io/clusterpedia/issues/168 needs to support wildcards and be able to collect all resources.
```yaml
spec:
  syncResources:
  - group: "*"
```
or
```yaml
spec:
  sycnResources:
  - group: "apps"
    resources: "*"
```
This feature requires clusterpedia to be able to dynamically discover changes in resource types in the cluster.

## When does a resource type change occur?
There are four cases:
1. kube-apiserver disables or enables a group/version or groups https://kubernetes.io/docs/tasks/administer-cluster/enable-disable-api/
2. kube upgrades result in changes to built-in resources, such as the addition or deprecation of a resource.
3. creation/deletion/update of a crd resource, resulting in a change of resource type
4.  creation/deletion of aggregated resources(APIServices), where the specific resources under the aggregated 

GroupVersion are controlled and provided by a specific third party component and are theoretically subject to dynamic change at will

For the third case, changes to the crd are actually easier to detect and we only need to listen for changes to the crd, so clusterpedia provides support for synchronising all custom resources first https://github.com/clusterpedia-io/clusterpedia/pull/112

The other three cases are actually a little more complex.

In the first case, we can watch changes to `APIService` to detect when group/version is disabled or enabled.

In the second case, kube's built-in resources will only change if there is a version change, so we need to monitor for version changes.

In the fourth case, the handling is less graceful than in the other cases. 
Firstly, the aggregated resource‘s Group/Version can also watch the `APIService` to get a coarse sense of changes to the aggregated resource at the GroupVersion level
For resource changes within versions, we need to poll regularly to monitor them.

## Some limitations of the restmapper

Firstly, the official RESTMapper(client-go) is a rather static implementation, we can usually use the [dynamic rest mapper](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/apiutil/dynamicrestmapper.go) provided by controller-runtime in order to get access to new resources in the cluster.

However, the dynamic rest mapper has some minor limitations. When a crd resource is deleted, the [dynamic rest mapper](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/apiutil/dynamicrestmapper.go) will not be aware of the deleted resource, but will still be able to get it through the rest mapper.

In addition, every time we look up a non-existent resource from the dynamic rest mapper, it will trigger a rebuild of the restmapper, requesting a lot of Discovery API, which is still a significant overhead despite the rebuild limit.

## How do we do it?
As clusterpedia is a multi-cluster tool, the additional losses in each cluster are multiplied by the number of clusters managed, so we need to try to avoid losses that can be avoided when implementing any of the features.

So how do we implement the monitoring of changes in cluster resource types?

Firstly, we didn't use a tool like dynamic restmapper, which could easily generate a lot of Discovery requests.

We should watch the `APIService` and `CRD`, so that we can detect changes in the kube apiserver's disabled resources and CR resource types.

In addition, by watching to these two resources, we can also distinguish which resource the group is: a *kube built-in resource*, a *custom resource*, or an *aggregated resource*.

By distinguishing between these three types of resources, when getting a resource type that is not recorded locally, we have four cases:
1. if the group is a *kube built-in resource*, and if the cluster version has not changed, then this resource type does not exist
2. if the group is a *custom resource*, which is not available locally because we are listening for changes to the CRD, then it does not exist
3. if the group is an *aggregated resource*, then we just need to refetch the resource information for the aggregated resource
4. if the group does not belong to any of the types, then it means that the resource does not exist

The only API requests in these four cases are to check if the cluster version has changed or to query for changes to resources under the aggregated resource’s group/version. This saves a lot of requests compared to the dynamic rest mapper rebuild

Even though we listen to APIService, we are still only aware of group and version changes, but we are still not able to actively detect kube version upgrades and intra-version resource changes due to aggregated resources

In fact the polling of cluster versions is done in order to handle seamless upgrades of cluster services without interruption, clusterpedia actually does a version check during the health check of the cluster from notready -> ready.
These two types of polling generally result in no more than 10 requests and can be relatively long in duration, which is very economical compared to the polling rebuild of the rest mapper


Iceber:
@ycsk02 @27149chen PTAL, Thanks

pkg/synchromanager/clustersynchro/discovery/discovery_cache.go
```diff
@@ -0,0 +1,446 @@
...
+	var hasList, hasWatch bool
+	for _, verb := range apiResource.Verbs {
+		switch verb {
+		case "list":
+			hasList = true
+		case "watch":
+			hasWatch = true
+		}
+	}
+	return hasList && hasWatch
```
27149chen, line 407: vs := sets.NewString(apiResource.Verbs)
return vs.HasAll("list", "watch")

pkg/synchromanager/clustersynchro/discovery/discovery_cache.go
```diff
@@ -0,0 +1,446 @@
...
+		syncResources = append(syncResources, resource)
+	}
+	return syncResources
+}
+
+func (c *DynamicDiscoveryManager) AttachAllCustomResourcesToSyncResources(resources []clusterv1alpha2.ClusterGroupResources) []clusterv1alpha2.ClusterGroupResources {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	sortedCustomResources := make([]schema.GroupResource, len(c.customResourceGroups))
```
27149chen, line 361: make([]schema.GroupResource, 0, len(c.customResourceGroups)) ?

pkg/synchromanager/clustersynchro/discovery/discovery_cache.go
```diff
@@ -0,0 +1,446 @@
...
+type GroupVersionCache interface {
+	SetGroupVersions(groupVersions map[string][]string, aggregatorGroups sets.String)
+}
+
+type CustomResourceCache interface {
+	UpdateCustomResource(resource schema.GroupResource, apiResource metav1.APIResource, versions []string)
+	RemoveCustomResource(resource schema.GroupResource)
+}
+
+func (c *DynamicDiscoveryManager) SetResourceMutationHandler(handler func()) {
```
27149chen, line 28: 1. 这个类型是否所有的导出方法都需要加锁，如果是，如何保证以后开发维护过程中不忘记加锁？
2. 通过是否以 locked 结尾来区分这个方法是否可以被用在加锁方法中，这样的机制不够清晰，容易误用

pkg/synchromanager/clustersynchro/apiservice_controller.go
```diff
@@ -0,0 +1,151 @@
...
+				if apiService.Spec.Service == nil {
+					// The versions of the group contain Local and Aggregator, ignoring such group
+					continue resolver
+				}
+
+				if !apiregistrationv1helper.IsAPIServiceConditionTrue(apiService, apiregistrationv1api.Available) {
+					// ignore unavailable aggregator version
+					continue
+				}
+			} else if apiService.Spec.Service != nil {
```
27149chen, line 132: 122，123，132 行都在判断 apiService.Spec.Service 是否是 nil，不是很理解意图是什么

pkg/synchromanager/clustersynchro/discovery/discovery_cache.go
Iceber: updated, This is more concise and clear

pkg/synchromanager/clustersynchro/discovery/discovery_cache.go
Iceber: fixed，thanks

pkg/synchromanager/clustersynchro/apiservice_controller.go
Iceber: 修改了一下逻辑，目的是为了保证相同 group 下的 apiservices 的类型一样，一个 group 不能同时包含 local 和 aggregator 两种类型
https://github.com/clusterpedia-io/clusterpedia/pull/180/files#diff-71110e8eb063ac8a48e782bc2050f440c43c4a2dfdc56e7ae1f6c13f65e71032R118

pkg/synchromanager/clustersynchro/discovery/discovery_cache.go
Iceber: 当前的规则是
* 对于导出方法是否加锁根据具体的函数功能来决定
* 对于非导出的方法，如果函数名包含了 locked 那么必须在调用上下文中加锁，以函数名称来人为约束

这种方式是借鉴了一下 client-go 的 DeltaFIFO，我也暂时没有想到其他更好的约束方式，有什么更好的建议吗

pkg/synchromanager/clustersynchro/apiservice_controller.go
27149chen: 这样看着清楚一些

pkg/synchromanager/clustersynchro/apiservice_controller.go
```diff
@@ -0,0 +1,150 @@
...
+
+		var versions []string
+		var isLocal, isAggregator bool
+		for _, apiService := range apiServicesByGroup {
+			if apiService.Spec.Service != nil {
+				isAggregator = true
+				if !apiregistrationv1helper.IsAPIServiceConditionTrue(apiService, apiregistrationv1api.Available) {
+					// ignore unavailable aggregator version
+					continue
+				}
```
ycsk02, line 126: The condition state has not yet been updated when a new aggregation API is created, and this check ignores the new APIService.

Iceber: Yes, it will ignore the abnormal aggregated APIService because we cannot fetch the resources of the abnormal aggregated api

ycsk02: When will the addition of the aggregate APIService Group to our discoveryCache be triggered.

Iceber: When the status of the `Available` condition changes to True, it usually means that the service can be accessed normally and will be added to the discoveryCache

pkg/synchromanager/clustersynchro/discovery/discovery_cache.go
```diff
@@ -0,0 +1,443 @@
...
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	sortedCustomResources := make([]schema.GroupResource, 0, len(c.customResourceGroups))
+	for gr := range c.resourceVersions {
+		if !c.customResourceGroups.Has(gr.Group) {
+			sortedCustomResources = append(sortedCustomResources, gr)
+		}
+	}
+	sortGroupResources(sortedCustomResources)
```
ycsk02, line 372: `!c.customResourceGroups.Has`  If it is included, it is a custom resource

Iceber: you are right

ycsk02:
Should we exclude Pediacluster from custom resources?  `HandleResourceEvent` always reports context Canceled Timeout error when Pediacluster resource synchronization is canceled by `synchro.cancel()` in `ResourceSynchro.Close` .

Iceber:
> Should we exclude Pediacluster from custom resources? HandleResourceEvent always reports context Canceled Timeout error when Pediacluster resource synchronization is canceled by synchro.cancel() in ResourceSynchro.Close 

After troubleshooting, I found that the error was wrapped in a layer in storage that caused the `errors.Is` to always return false, so even if it was due to context closure it would still cause the log to print.

https://github.com/clusterpedia-io/clusterpedia/blob/85f7337d510656df0b57a0b751a12bf4a564f74e/pkg/synchromanager/clustersynchro/resource_synchro.go#L344-L352

should return `err` directly
https://github.com/clusterpedia-io/clusterpedia/blob/85f7337d510656df0b57a0b751a12bf4a564f74e/pkg/storage/internalstorage/errors.go#L42-L43

ycsk02:
LGTM

Issue #177, opened 2022-05-07
Support for setting feature gates in helm charts


Iceber:
### What would you like to be added?

Add settings for feature gates in *values.yaml*, `apiserver` and `clustersynchro manager` need to be able to be set separately.

### Why is this needed?

We need to support the setting of feature gates when deploying with helm, otherwise it's a pain to modify them after the deployment is complete.

PR #181, opened 2022-05-11
add helm featureGates

calvin0327:
Signed-off-by: calvin0327 <wen.chen@daocloud.io>

/kind feature

fixed: https://github.com/clusterpedia-io/clusterpedia/issues/177

add helm featureGates

### DIFF
--- a/charts/values.yaml
+++ b/charts/values.yaml
@@ -41,6 +41,7 @@ commonAnnotations: {}
 ## if PV resources are not required, set it to None!
 ## if defined, will create the pv on the node.
 persistenceMatchNode: ""
+
 ## @param config of clusterpedia log
 log:
   ## @param enabled indicates whether enable log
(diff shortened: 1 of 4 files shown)


charts/templates/_common.tpl
```diff
@@ -0,0 +1,25 @@
+{{- define "common.featureGate" -}}
```
Iceber, line 1: Usually apiserver and clustersynchro have different feature gates, so it is recommended to keep them separate

charts/values.yaml
```diff
@@ -180,6 +192,12 @@ clustersynchroManager:
   tolerations:
     - key: node-role.kubernetes.io/master
       operator: Exists
+  ## @param featureGate to clustersynchro
+  featureGates:
+    - name: PruneManagedFields
+      enabled: true
+    - name: PruneLastAppliedConfiguration
+      enabled: true
```
Iceber, line 200: I think it might be better to use the map form.
eg.
```yaml
  featureGates:
    PruneManagedFields: true
    PruneLastAppliedConfiguration: true
```
This way we can use `--set clustersynchroManager.featurGates.PruneManagedFields: true` to set it when installing or upgrading

Iceber, line 200: I think it might be better to use the map form.
eg.
```yaml
  featureGates:
    PruneManagedFields: true
    PruneLastAppliedConfiguration: true
```
This way we can use `--set clustersynchroManager.featurGates.PruneManagedFields: true` to set it when installing or upgrading

Iceber (approved):
LGTM

Iceber:
A good solution to the problem of default values for feature gates when deploying. 
https://github.com/clusterpedia-io/clusterpedia/issues/81#issuecomment-1120124783

Issue #13, opened 2021-12-06
日志最好有轮滚，避免出现把磁盘写满的情况


yank1:
参考

https://github.com/clusterpedia-io/clusterpedia/blob/7e301c8ce4cf81827850ab0f58f64b0457563f76/pkg/storage/internalstorage/register.go#L87

Iceber:
Although the `internalstorage` will only print logs of errors and Slow SQL by default, there is still a risk of oversizing the log file in some cases.
It will be added to the roadmap

PR #182, opened 2022-05-16
Support log rotate

RuliXu:
support: https://github.com/clusterpedia-io/clusterpedia/issues/13

### DIFF
--- a/pkg/storage/internalstorage/config.go
+++ b/pkg/storage/internalstorage/config.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/go-sql-driver/mysql"
 	"github.com/jackc/pgx/v4"
+	"gopkg.in/natefinch/lumberjack.v2"
 	"gorm.io/gorm/logger"
 )
 
@@ -39,11 +40,12 @@ type Config struct {
 }
 
 type LogConfig struct {
-	Stdout                    bool          `yaml:"stdout"`
-	Level                     string        `yaml:"level"`
-	Colorful                  bool          `yaml:"colorful"`
-	SlowThreshold             time.Duration `yaml:"slowThreshold" default:"200ms"`
-	IgnoreRecordNotFoundError bool          `yaml:"ignoreRecordNotFoundError"`
+	Stdout                    bool               `yaml:"stdout"`
+	Level                     string             `yaml:"level"`
+	Colorful                  bool               `yaml:"colorful"`
+	SlowThreshold             time.Duration      `yaml:"slowThreshold" default:"200ms"`
+	IgnoreRecordNotFoundError bool               `yaml:"ignoreRecordNotFoundError"`
+	Logger                    *lumberjack.Logger `yaml:"logger"`
 }
 
 type MySQLConfig struct {
--- a/pkg/storage/internalstorage/register.go
+++ b/pkg/storage/internalstorage/register.go
@@ -10,6 +10,7 @@ import (
 	"github.com/go-sql-driver/mysql"
 	"github.com/jackc/pgx/v4/stdlib"
 	"github.com/jinzhu/configor"
+	"gopkg.in/natefinch/lumberjack.v2"
 	gmysql "gorm.io/driver/mysql"
 	gpostgres "gorm.io/driver/postgres"
 	"gorm.io/gorm"
@@ -19,6 +20,7 @@ import (
 )
 
 const StorageName = "internal"
+const defaultLogFileName = "/var/log/clusterpedia/internalstorage.log"
 
 func init() {
 	storage.RegisterStorageFactoryFunc(StorageName, NewStorageFactory)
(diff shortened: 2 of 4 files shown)


pkg/storage/internalstorage/register.go
```diff
@@ -90,9 +91,17 @@ func newLogger(cfg *Config) (logger.Interface, error) {
...
 		}
 
-		logWriter, err = os.OpenFile("/var/log/clusterpedia/internalstorage.log", os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0644)
+		path := "/var/log/clusterpedia/internalstorage.log"
+		f, err := os.OpenFile(path, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0644)
 		if err != nil {
-			return nil, fmt.Errorf("open storage log file error: %w", err)
+			return nil, fmt.Errorf("open storage log file %s error: %w", path, err)
+		}
+		f.Close()
```
cleverhu, line 99: Data cannot be written after the file is closed.

pkg/storage/internalstorage/config.go
```diff
@@ -35,7 +35,10 @@ type Config struct {
 
 	Params map[string]string `yaml:"params"`
 
-	Log *LogConfig `yaml:"log"`
+	Log        *LogConfig `yaml:"log"`
+	MaxBackups int        `yaml:"maxBackups"`
```
cleverhu, line 39: These new parameters are all log parameters. They should be placed in the log so that users can configure them.

pkg/storage/internalstorage/register.go
```diff
@@ -90,9 +91,17 @@ func newLogger(cfg *Config) (logger.Interface, error) {
 			return nil, fmt.Errorf("mkdir log dir error: %w", err)
 		}
 
-		logWriter, err = os.OpenFile("/var/log/clusterpedia/internalstorage.log", os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0644)
+		path := "/var/log/clusterpedia/internalstorage.log"
```
cleverhu, line 94: The path should not be fixed. It should have a default value or be taken from yaml. In addition, consider what happens if the user doesn't want to keep the log and outputs it directly to the terminal.

pkg/storage/internalstorage/register.go
RuliXu: > Data cannot be written after the file is closed.

This is only an ensureFile logic, lumberjack.Logger would open file of path later.

pkg/storage/internalstorage/register.go
RuliXu: > The path should not be fixed. It should have a default value or be taken from yaml. 

I will make it flexible.
> In addition, consider what happens if the user doesn't want to keep the log and outputs it directly to the terminal.

We have the Stdout option now.

pkg/storage/internalstorage/config.go
```diff
@@ -44,6 +44,12 @@ type LogConfig struct {
 	Colorful                  bool          `yaml:"colorful"`
 	SlowThreshold             time.Duration `yaml:"slowThreshold" default:"200ms"`
 	IgnoreRecordNotFoundError bool          `yaml:"ignoreRecordNotFoundError"`
+	Filename                  string        `yaml:"filename"`
```
cleverhu, line 47: Will it be better to use these parameters in the nested structure?

RuliXu: Great idea

pkg/storage/internalstorage/config.go
```diff
@@ -39,11 +40,12 @@ type Config struct {
...
-	Level                     string        `yaml:"level"`
-	Colorful                  bool          `yaml:"colorful"`
-	SlowThreshold             time.Duration `yaml:"slowThreshold" default:"200ms"`
-	IgnoreRecordNotFoundError bool          `yaml:"ignoreRecordNotFoundError"`
+	Stdout                    bool              `yaml:"stdout"`
+	Level                     string            `yaml:"level"`
+	Colorful                  bool              `yaml:"colorful"`
+	SlowThreshold             time.Duration     `yaml:"slowThreshold" default:"200ms"`
+	IgnoreRecordNotFoundError bool              `yaml:"ignoreRecordNotFoundError"`
+	Logger                    lumberjack.Logger `yaml:"logger"`
```
cleverhu, line 48: This should be a pointer because it is optional.

pkg/storage/internalstorage/register.go
Iceber: lumberjack will create `dir` and `filename`, this should be a capability of lumberjack, we don't need to create additional

pkg/storage/internalstorage/register.go
```diff
@@ -86,14 +89,28 @@ func newLogger(cfg *Config) (logger.Interface, error) {
 	if cfg.Log.Stdout {
 		logWriter = os.Stdout
 	} else {
-		if err := os.MkdirAll("/var/log/clusterpedia", 0644); err != nil {
-			return nil, fmt.Errorf("mkdir log dir error: %w", err)
+		loggerConf := cfg.Log.Logger
```
Iceber, line 92: `loggerConf` doesn't look clear, and it is not a config. `lumberjackLogger` looks better than `loggerConfig`.

Iceber (approved):
LGTM，Thank you very much for contributing, this is a very useful feature

RuliXu:
> LGTM，Thank you very much for contributing, this is a very useful feature

Thank you for your help!

PR #184, opened 2022-05-19
add ClusterSyncResources

cleverhu:
How to use:
you can create the crd by yaml
```
apiVersion: cluster.clusterpedia.io/v1alpha2
kind: ClusterSyncResources
metadata:
  name: cluster-sync-resources-example
spec:
  syncResources:
    - group: ""
      resources:
        - pods
      versions:
        - v1
````
and edit pediacluster.spec.ClusterSyncResources: cluster-sync-resources-example, then you will see it works!

resolve: https://github.com/clusterpedia-io/clusterpedia/issues/150

### CODE REVIEW

File: pkg/synchromanager/clustersynchro_manager.go [hunk 1]
```diff
@@ -341,6 +362,71 @@ func (manager *Manager) UpdateClusterStatus(ctx context.Context, name string, st
...
+	template := new.(*clusterv1alpha2.ClusterSyncResourcesTemplate)
+	clusters, err := manager.clusterlister.List(labels.Everything())
+	if err != nil {
+		klog.Error("list clusters failed while updating template: %v", err)
+		return
+	}
+	for _, cluster := range clusters {
+		if cluster.Spec.SyncResourcesTemplate == template.Name {
+			manager.enqueue(cluster)
+			return
```
Iceber, line 390: Need to queue all clusters that reference the template

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -341,6 +362,71 @@ func (manager *Manager) UpdateClusterStatus(ctx context.Context, name string, st
...
+	template := obj.(*clusterv1alpha2.ClusterSyncResourcesTemplate)
+	clusters, err := manager.clusterlister.List(labels.Everything())
+	if err != nil {
+		klog.Error("list clusters failed while updating template : %v", err)
+		return
+	}
+	for _, cluster := range clusters {
+		if cluster.Spec.SyncResourcesTemplate == template.Name {
+			manager.enqueue(cluster)
+			return
```
Iceber, line 375: Need to queue all clusters that reference the template

cleverhu: done

calvin0327:
Shall we start directly from the v1alpha2 version?

Iceber:
> Shall we start directly from the v1alpha2 version?

When the stable version is not considered, new resources can be added directly to the latest version(alpha/beta)

Iceber:
For template deletion we may need to be a bit stricter.

If there are still clusters referencing the template, the template cannot be deleted ？

calvin0327:
The relationship between template and pediaCluster should be undone.  the template is only work in creating pediacluster.
I think so, Otherwise it becomes very complicated.

Iceber:
> The relationship between template and pediaCluster should be undone. the template is only work in creating pediacluster.

`sync resource templates` do not only have an effect on creation, future updates to the resource templates will also result in changes to the pediacluster synchronized resources, what do you think?

RuliXu:
Might need to rebase main after https://github.com/clusterpedia-io/clusterpedia/pull/221 is merged.

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -282,8 +292,18 @@ func (manager *Manager) reconcileCluster(cluster *clusterv1alpha2.PediaCluster)
...
 
 	manager.UpdateClusterSynchroCondition(cluster.Name, clusterv1alpha2.RunningConditionReason, "", metav1.ConditionTrue)
+	syncResources := cluster.Spec.SyncResources
+	if templateName := cluster.Spec.ClusterSyncResources; templateName != "" {
+		template, err := manager.clusterSyncResourcesLister.Get(templateName)
+		if err != nil && !apierrors.IsNotFound(err) {
+			klog.ErrorS(err, "Failed to get cluster sync template", "cluster", cluster.Name)
+		}
+		if template != nil {
+			syncResources = mergeResources(syncResources, template.Spec.SyncResources)
```
Iceber, line 302: Instead of using map to de-duplicate, just append the two slices together and clustersynchro will take care of the de-duplication.

```go
syncResources = append(syncResources, template.Spec.SyncResources)
```

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -282,8 +292,18 @@ func (manager *Manager) reconcileCluster(cluster *clusterv1alpha2.PediaCluster)
 	}
 
 	manager.UpdateClusterSynchroCondition(cluster.Name, clusterv1alpha2.RunningConditionReason, "", metav1.ConditionTrue)
+	syncResources := cluster.Spec.SyncResources
+	if templateName := cluster.Spec.ClusterSyncResources; templateName != "" {
+		template, err := manager.clusterSyncResourcesLister.Get(templateName)
+		if err != nil && !apierrors.IsNotFound(err) {
+			klog.ErrorS(err, "Failed to get cluster sync template", "cluster", cluster.Name)
```
Iceber, line 299: Usually Error means that the program is abnormal and needs to end the subsequent logic

It is recommended to use `Warn` and add the TODO that we discussed last time to do in the future

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -282,8 +292,18 @@ func (manager *Manager) reconcileCluster(cluster *clusterv1alpha2.PediaCluster)
 	}
 
 	manager.UpdateClusterSynchroCondition(cluster.Name, clusterv1alpha2.RunningConditionReason, "", metav1.ConditionTrue)
+	syncResources := cluster.Spec.SyncResources
+	if templateName := cluster.Spec.ClusterSyncResources; templateName != "" {
+		template, err := manager.clusterSyncResourcesLister.Get(templateName)
```
Iceber, line 297: s/ClusterSyncResources/SyncResourcesRefName
s/template/ref

Iceber:
@RuliXu The first commit doesn't seem to make sense and can be rebased into a single commit

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -357,6 +378,34 @@ func (manager *Manager) UpdateClusterStatus(ctx context.Context, name string, st
...
+	}
+	for _, cluster := range clusters {
+		if cluster.Spec.SyncResourcesRefName == template.Name {
+			manager.enqueue(cluster)
+		}
+	}
+}
+
+func (manager *Manager) updateClusterSyncResources(old, new interface{}) {
+	template := new.(*clusterv1alpha2.ClusterSyncResources)
```
Iceber, line 396: s/template/ref

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -357,6 +378,34 @@ func (manager *Manager) UpdateClusterStatus(ctx context.Context, name string, st
 	})
 }
 
+func (manager *Manager) addOrDeleteClusterSyncResources(obj interface{}) {
+	template := obj.(*clusterv1alpha2.ClusterSyncResources)
+	clusters, err := manager.clusterlister.List(labels.Everything())
+	if err != nil {
+		klog.Error("list clusters failed while updating template : %v", err)
```
Iceber, line 385:
```go
klog.ErrorS(err, "list clusters failed while adding/deleting clustersyncresources", "clustersyncresources", ref.Name)
```

Iceber (commented):
`addOrDeleteCluster` and `updateClusterSyncResources` look logically the same

```go
	clusterSyncResourcesInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
		AddFunc:    manager.handleClusterSyncResources,
		UpdateFunc: func(_, obj interface{}){manager.handleClusterSyncResources(obj)},
		DeleteFunc: manager.handleClusterSyncResources,
	})


func (manager *Manager) handleClusterSyncResources( obj interface{}) {
    // ClusterSyncResources is cluster scoped resource, key is name
    refName, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
    if err != nil {
    ...
    }
    // other logic
}

File: pkg/synchromanager/clustersynchro_manager.go [hunk 8]
```diff
@@ -282,8 +292,19 @@ func (manager *Manager) reconcileCluster(cluster *clusterv1alpha2.PediaCluster)
...
+	syncResources := cluster.Spec.SyncResources
+	if refName := cluster.Spec.SyncResourcesRefName; refName != "" {
+		ref, err := manager.clusterSyncResourcesLister.Get(refName)
+		if err != nil && !apierrors.IsNotFound(err) {
+			// TODO: Use more obvious method to let users know.
+			klog.Warning(err, "Failed to get SyncResourcesRef of cluster", cluster.Name)
+		}
+		if ref != nil {
+			syncResources = append(syncResources, ref.Spec.SyncResources...)
+		}
```
Iceber, line 304: if err != nil && !apierrors.IsNotFound(err)，we should return err

if (err == nil && apierrors.IsNofFound(err)) || ref == nil, we should warning.

Iceber (approved):
LGTM

PR #188, opened 2022-05-25
support set sync resources with wildcard

Iceber:
issue: https://github.com/clusterpedia-io/clusterpedia/issues/168

all-resource wildcards
```yaml
- group: "*"
   resources:
   - "*"
```
 group resource wildcards
```yaml
- group: "apps"
   resources:
   - "*"
```

Note:
The all-resource wildcard will generate many long connections, so if the user does not need to synchronize all resources, it is recommended to specify the type of resources or only synchronize custom resources https://github.com/clusterpedia-io/clusterpedia/pull/112.

And if you want to use all-resource wildcard, you need to enable feature gate `--feature-gates=AllowSyncAllResources=true`.

### DIFF
--- a/charts/values.yaml
+++ b/charts/values.yaml
@@ -196,15 +196,22 @@ clustersynchroManager:
     ## alpha: v0.0.9
     ## beta: v0.3.0
     PruneManagedFields: true
+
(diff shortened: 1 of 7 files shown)


calvin0327:
@Iceber the `values.yaml `of chart pkg  need update yet.

Iceber:
@calvin0327 Thanks, Updated

PR #191, opened 2022-05-26
fix chart featureGate

calvin0327:

/kind bug

Fixed the  judge condition of featureGate value. 
The featureGate value set by `--set` may be a bool or a string.  only it's a empty string do we not to render the featureGate item to yaml template files.  if the value is false, it should alse be rendered to yaml files. 

wo should switch the bool value to string and judge its length.


calvin0327:
@Iceber PTAL

Iceber (approved):
LGTM

Issue #196, opened 2022-06-01
The feature gate `RemainingItemCount` should default to false


Iceber:
### What happened?

https://github.com/clusterpedia-io/clusterpedia/issues/81  uses kube's native feature gate `RemainingItemCount` to set whether RemainingItemCount is returned by default.
```bash
$ ./bin/apiserver --help | grep RemainingItemCount
                                      RemainingItemCount=true|false (BETA - default=true)
```

If the feature gate is not explicitly set to true, then every search will return the remaining count, adding additional sql queries to the default storage layer

### What did you expect to happen?

The default value of this feature gate is True, and there is no need for clusterpedia to return RemainingItemCount by default, which may cause unnecessary performance loss.
We always need to actively set it to false.
```bash
./bin/apiserver --feature-gates RemainingItemCount=false
```

So we want to make this feature gate default to false, so that the user does not have to set the feature gate RemainingItemCount every time.

### How can we reproduce it (as minimally and precisely as possible)?

The feature gate `RemainingItemCount` is not explicitly set to true.

### Clusterpedia Version

<details>

```console
# replace the pod name of clusterpedia apiserver
$ kubectl -n clusterpedia-system exec {clusterpedia-apiserver-pod-name} -- apiserver --version=raw
# paste output here

# replace the pod name of clusterpedia clustersynchro-manager
$ kubectl -n clusterpedia-system exec {clustersynchro-manager-pod-name} -- clustersynchro-manager --version=raw
# paste output here
```

</details>

### Host Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

PR #197, opened 2022-06-01
Change the default value of the feature gate RemainintItemCount to false

cleverhu:
/resolve: https://github.com/clusterpedia-io/clusterpedia/issues/196


Iceber (approved):
LGTM

PR #200, opened 2022-06-01
Update Helm Dependency

RuliXu:
Update helm chart dependency.


calvin0327:
/LGTM

upgrate  `chart.yaml` version to sub chart only update chart pkg.

PR #203, opened 2022-06-06
fix collection resources order

cleverhu:
When the resource is a map, the order of `kubectl get collectionresource` cannot be guaranteed after the apiserver is restarted.


Iceber (approved):
LGTM

PR #221, opened 2022-06-10
Bump golangci lint to v1.46.2

RuliXu:


Iceber:
golangci-lint often times out, we can increase the timeout to 10m
https://github.com/clusterpedia-io/clusterpedia/blob/19a5b23b61ed19899b68420b70071d33cbdd14da/.golangci.yml#L1-L3

Iceber (approved):
LGTM

PR #225, opened 2022-06-12
bug fix: os.exit will not wait for the defer function

hanweisen:
FlushLogs()will not call if os.Exit(1)
referring to 
https://github.com/karmada-io/karmada/pull/1289
https://github.com/kubernetes/kubernetes/pull/104503

### DIFF
--- a/cmd/apiserver/main.go
+++ b/cmd/apiserver/main.go
@@ -11,12 +11,19 @@ import (
 )
 
 func main() {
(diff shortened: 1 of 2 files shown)


hanweisen:
@Iceber

Iceber (approved):
LGTM, Thank you very much for fixing this bug!🎉

Iceber:
In the future we are going to use `k8s.io/component-base/cli` to run the command

PR #233, opened 2022-06-15
add exception message

qiuming520:

### DIFF
--- a/hack/install-golangci-lint.sh
+++ b/hack/install-golangci-lint.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# do proxy setting in China mainland
+if [[ -n ${CHINA_MAINLAND:-} ]]; then
+  export GOPROXY=https://proxy.golang.com.cn,direct # set domestic go proxy
+fi
+echo "install golangci-lint"
+GO111MODULE=on go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.46.2
--- a/pkg/apiserver/apiserver.go
+++ b/pkg/apiserver/apiserver.go
@@ -2,7 +2,7 @@ package apiserver
 
 import (
 	"context"
-	"errors"
+	"fmt"
 	"net/http"
 
 	metainternal "k8s.io/apimachinery/pkg/apis/meta/internalversion"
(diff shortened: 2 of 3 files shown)


pkg/apiserver/apiserver.go
```diff
@@ -100,10 +100,10 @@ func (cfg *Config) Complete() CompletedConfig {
 
 func (config completedConfig) New() (*ClusterPediaServer, error) {
 	if config.ClientConfig == nil {
-		return nil, errors.New("")
+		return nil, errors.New("ClientConfig is missing")
 	}
 	if config.StorageFactory == nil {
-		return nil, errors.New("")
+		return nil, errors.New("StorageFactory config is missing")
```
Iceber, line 106: We could use the same error style as `genericapiserver`, after all, they are functions with the same meaning.
https://github.com/clusterpedia-io/clusterpedia/blob/01ad71bfa087b18e16eb6c981c874c023c91acff/vendor/k8s.io/apiserver/pkg/server/config.go#L592-L601

eg. `apiserver.New() called with config.ClientConfig == nil`
> `Clusterpediaapiserver.New` and  `APIServer.New` don't look good, we can just use the `apiserver.New`

Iceber (commented):
Thanks for the fix, just the error message looks like it could be slightly optimized

pkg/apiserver/apiserver.go
qiuming520: I have revised it as suggested .

Iceber:
LGTM, Please rebase commits into one, because you are actually doing one thing.
> In order to try to merge in a uniform way, we don't use `squash and merge`

hack/install-golangci-lint.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# do proxy setting in China mainland
+if [[ -n ${CHINA_MAINLAND:-} ]]; then
+  export GOPROXY=https://proxy.golang.com.cn,direct # set domestic go proxy
+fi
```
Iceber, line 10: In China, the developers usually configure this environment variable, and it should not be configured by the script, after all, maybe the user has their own defined `GOPROXY` that cannot override the user environment's proxy.
What do you think?

qiuming520: yeah！ Your point is right. I have revoked this PR
Just it's easy to ignore proxy settings when installing in a new environment.

PR #239, opened 2022-06-16
test: add check dependencies

qiuming520:
check whether 'kind' is installed

### DIFF
--- a/test/helper.sh
+++ b/test/helper.sh
@@ -1,6 +1,30 @@
 #!/usr/bin/env bash
-
 ROOT="$(dirname "${BASH_SOURCE[0]}")/.."
+
+# check whether command is installed.
+function cmd_exist() {
+    local command="${1}"
+    type "${command}" >/dev/null 2>&1
+}
+
+# Check dependencies is installed or not and exit if not
+function check_dependencies() {
+    local dependencies=("${@}")
+    local not_installed=()
+    for dependency in "${dependencies[@]}"; do
+        if ! cmd_exist "${dependency}"; then
+            not_installed+=("${dependency}")
+        fi
+    done
+
+    if [[ "${#not_installed[@]}" -ne 0 ]]; then
+        echo "Error: Some dependencies are not installed:"
+        for dependency in "${not_installed[@]}"; do
+            echo "  - ${dependency}"
+        done
+        exit 1
+    fi
+}
 
 # build the image for the test
 function build_image() {
--- a/test/test.sh
+++ b/test/test.sh
@@ -9,6 +9,10 @@ function main() {
     local env_name
     local failed=()
 
+    echo "::group::check dependencies"
+    check_dependencies docker kind kubectl helm
+    echo "::endgroup::"
+
     echo "::group::Build image"
     build_image
     echo "::endgroup::"


Iceber:
@wzshiming PTAL, Thanks

test/helper.sh
```diff
@@ -1,7 +1,20 @@
...
 
+# check whether command is installed.
+function cmd_must_exist {
+    local CMD=$(command -v ${1})
+    if [[ ! -x ${CMD} ]]; then
+      echo "error: Please install ${1} and verify they are in \$PATH."
+      exit 1
+    fi
+}
+
```
wzshiming, line 17:
```suggestion
# check whether command is installed.
function cmd_must_exist {
    if ! command -v ${1} > /dev/null 2>&1; then
      echo "error: Please install ${1} and verify they are in \$PATH."
      exit 1
    fi
}

```

For builtin and alias commands, use -x will judgment fails. 
We could just use `command -v`.

test/helper.sh
```diff
@@ -14,6 +27,8 @@ function load_image() {
     if ! docker image inspect "${image}" >/dev/null 2>&1; then
         docker pull "${image}"
     fi
+
+    cmd_must_exist "kind"
```
wzshiming, line 31: Maybe it's better to add all dependency checks at the beginning of the test, rather than at every location.

wzshiming (changes requested):

wzshiming:
``` bash
# Check a command exist
function command_exist() {
    local command="${1}"
    type "${command}" >/dev/null 2>&1
}

# Check dependencies is installed or not and exit if not
function check_dependencies() {
    local dependencies=("${@}")
    local not_installed=()
    for dependency in "${dependencies[@]}"; do
        if ! command_exist "${dependency}"; then
            not_installed+=("${dependency}")
        fi
    done

    if [ "${#not_installed[@]}" -ne 0 ]; then
        echo "Error: Some dependencies are not installed:"
        for dependency in "${not_installed[@]}"; do
            echo "  - ${dependency}"
        done
        exit 1
    fi
}


check_dependencies docker kind kubectl
```

test/helper.sh
```diff
@@ -1,7 +1,20 @@
 #!/usr/bin/env bash
 
+set -o errexit
```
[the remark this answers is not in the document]
qiuming520: remove  ‘set -o errexit’

test/helper.sh
```diff
@@ -1,7 +1,20 @@
...
+set -o nounset
+set -o pipefail
```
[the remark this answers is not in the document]
qiuming520: have removed

test/helper.sh
qiuming520: add 
# check dependency
cmd_must_exist "kind"
cmd_must_exist "docker"
cmd_must_exist "kubectl"
cmd_must_exist "helm"

wzshiming:
https://github.com/clusterpedia-io/clusterpedia/blob/9ef1a04b28763b4ff5a02f7cd1967b5688f28c8a/test/test.sh#L7-L15

This L11 is a good place to check dependencies.

test/helper.sh
```diff
@@ -1,7 +1,31 @@
 #!/usr/bin/env bash
-
 ROOT="$(dirname "${BASH_SOURCE[0]}")/.."
 
+# check whether command is installed.
+function cmd_must_exist() {
```
wzshiming, line 5:
```suggestion
function cmd_exist() {
```

wzshiming (commented):
Others lgtm

wzshiming (approved):
/assign @Iceber

lgtm for me

Iceber (approved):
LGTM, Thanks to @qiuming520  for contributing and @wzshiming  for reviewing

Iceber:
@qiuming520 The commit message is too long, suggest to change it to `test: check dependencies` or other more concise message

qiuming520:
> Here https://github.com/clusterpedia-io/clusterpedia/pull/239/commits also needs to be updated, the commit message is already long enough to be truncated

yeah ！ I have changed~ @Iceber

PR #244, opened 2022-06-23
Optimize：add description for cmd package _ import and optimize error log

qiuming520:

### DIFF
--- a/pkg/apiserver/registry/clusterpedia/collectionresources/rest.go
+++ b/pkg/apiserver/registry/clusterpedia/collectionresources/rest.go
@@ -2,14 +2,15 @@ package collectionresources
 
 import (
 	"context"
-	"errors"
 	"strings"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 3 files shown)


pkg/apiserver/registry/clusterpedia/collectionresources/rest.go
```diff
@@ -109,7 +109,7 @@ func (s *REST) Get(ctx context.Context, name string, _ *metav1.GetOptions) (runt
 
 	storage, ok := s.storages[name]
 	if !ok {
-		return nil, errors.New("")
+		return nil, fmt.Errorf("CollectionResourceStorage failed to get key: %v\n", name)
```
Iceber, line 112: Thanks for fixing this legacy issue from alpha.

I think this might be better, otherwise it would be `500 Internal Server Error`.
```go
                #  import apierrors "k8s.io/apimachinery/pkg/api/errors"

                return nil, apierrors.NewNotFound(
                        schema.GroupResource{Group: internal.GroupName, Resource: "collectionresources"},
                        name,
                )
```

qiuming520: Thanks for your suggestion !  I have changed

Iceber (approved):
LGTM

Issue #238, opened 2022-06-16
cmd/clustersynchro-manager has duplicated log tags


Iceber:
### What happened?

duplicate log tags
```bash
./bin/clustersynchro-manager --help
...
Logs flags:

      --log-flush-frequency duration
                Maximum number of seconds between log flushes (default 5s)
      --logging-format string
                Sets the log format. Permitted formats: "text".
                Non-default formats don't honor these flags: --add-dir-header, --alsologtostderr, --log-backtrace-at, --log-dir, --log-file, --log-file-max-size, --logtostderr, --one-output,
                --skip-headers, --skip-log-headers, --stderrthreshold, --vmodule.
                Non-default choices are currently alpha and subject to change without warning. (default "text")
  -v, --v Level
                number for the log level verbosity
      --vmodule pattern=N,...
                comma-separated list of pattern=N settings for file-filtered logging (only works for text log format)
Global flags:
...
  -v, --v Level
                number for the log level verbosity
      --vmodule moduleSpec
                comma-separated list of pattern=N settings for file-filtered logging
```

### What did you expect to happen?

Only need to set log tags once

### How can we reproduce it (as minimally and precisely as possible)?

```bash
$ make clustersynchro-manager
$ ./bin/clustersynchro-manager --help
```

### Anything else we need to know?

we need to add `logs.SkipLoggingConfigurationFlags()` to the globalflag.AddGlobalFlags to prevent duplicate registration of log flags.
https://github.com/clusterpedia-io/clusterpedia/blob/f131cccc4e2b5953134fb47236e8d74898cd73c1/cmd/clustersynchro-manager/app/synchro.go#L70
to
```
 globalflag.AddGlobalFlags(namedFlagSets.FlagSet("global"), cmd.Name(),logs.SkipLoggingConfigurationFlags()) 
```

### Clusterpedia Version

<details>

```console
# replace the pod name of clusterpedia apiserver
$ kubectl -n clusterpedia-system exec {clusterpedia-apiserver-pod-name} -- apiserver --version=raw
# paste output here

# replace the pod name of clusterpedia clustersynchro-manager
$ kubectl -n clusterpedia-system exec {clustersynchro-manager-pod-name} -- clustersynchro-manager --version=raw
# paste output here
```

</details>

### Host Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

PR #246, opened 2022-06-24
fix:cmd/clustersynchro-manager has duplicated log tags

ONE7live:
Signed-off-by: ONE7live <wangqi_yewu@cmss.chinamobile.com> 

#238 

fix:cmd/clustersynchro-manager has duplicated log tags.


Iceber (approved):
LGTM, Thanks for you contribution🎉

PR #269, opened 2022-07-19
fix chart role name

calvin0327:

/kind bug

If the chart as a sub chart, `Release.name` is very easy to conflict with parent chart. we should try not to use 'release.name' directly.


Iceber (approved):
LGTM

Iceber:
We may need to refine the rules needed for components in the future

PR #270, opened 2022-07-19
fix storage images registry

calvin0327:

/kind bug

The storage (postgresql & mysql) is out of `global.imaesRegistry` control.


calvin0327:
@Iceber I have a test, It still makes sense to mysql and postgresql.

calvin0327:
https://github.com/clusterpedia-io/clusterpedia/pull/312

Has been solved.

PR #304, opened 2022-08-09
unify feature gate in aa、sync-controller and cluster-controller

wuyingjun-lucky:

**What type of PR is this?**
/kind cleanup

unify default feature gate in aa、sync-controller and cluster-controller
Fixes #

```release-note
NONE
```


Iceber:
The reason for separating `clustersynchro manager`'s feature gate from `clusterpedia apiserver`'s is that apiserver's doesn't make any sense for `clustersynchro manager`, so a separate feature gate is used for clustersynchro manager in order to get a better view of the useful feature-gate

wuyingjun-lucky:
> The reason for separating `clustersynchro manager`'s feature gate from `clusterpedia apiserver`'s is that apiserver's feature gates doesn't make any sense for `clustersynchro manager`, so a separate feature gate is used for clustersynchro manager in order to get a better view of the useful feature-gate

I am lack of knowledge here.  What I am confused about is two points：
1· cluster controller use the same feature-gates as sync controller 
2. AA and sync-controller start in two containers . Does that make have an effect when we do not separate the feature_gates
Can you help me?

Iceber:
> 1. cluster controller use the same feature-gates as sync controller

They just use the same utils/feature package, which is actually empty if we don't import the specific features package.

[utils/feature](https://github.com/clusterpedia-io/clusterpedia/tree/main/pkg/utils/feature) can be used as a public feature gate for clusterpedia, and that components will register its own features if it needs to use them.


> 2. AA and sync-controller start in two containers . Does that has an effect when we do not separate the feature_gates

Indeed, if the command is integrated, both feature gates will be used.

There are two ways to do this, which are 
* Register all features in DefaultMutableFeatureGate, but for the other two components, there is no way to see the really useful feature gate with --help

* Add a `SetMutableFeatureGate` function in *utils/features* to override MutableFeatureGate in *utils/features*, and add an `AddFeatureGates` function in *pkg/synchromanager/features*, like [logs.AddFeatureGates](https://github.com/clusterpedia-io/clusterpedia/blob/main/cmd/clustersynchro-manager/app/synchro.go#L28)

PR #308, opened 2022-08-10
add go version minimum requirement detection

ONE7live:
Signed-off-by: ONE7live <wangqi_yewu@cmss.chinamobile.com>

**What type of PR is this?**
/kind feature

I had a problem when I started clusterpedia with a lower version of go, so I would like to suggest setting a minimum go version and check before starting.

Fixes #

```release-note
NONE
```

### DIFF
--- a/hack/update-vendor.sh
+++ b/hack/update-vendor.sh
@@ -7,6 +7,21 @@ set -o pipefail
 REPO_ROOT=$(git rev-parse --show-toplevel)
 cd "${REPO_ROOT}"
 
+MIN_GO_VERSION=go1.17.0
+
+# Ensure the go tool exists and is a viable version.
+function verify_go_version {
+  local GO_VERSION
+  IFS=" " read -ra GO_VERSION <<< "$(GOFLAGS='' go version)"
+      if [[ "${MIN_GO_VERSION}" != $(echo -e "${MIN_GO_VERSION}\n${GO_VERSION[2]}" | sort -s -t. -k 1,1 -k 2,2n -k 3,3n | head -n1) && "${GO_VERSION[2]}" != "devel" ]]; then
+        echo "Detected go version: ${GO_VERSION[*]}."
+        echo "Clusterpedia requires ${MIN_GO_VERSION} or greater."
+        echo "Please install ${MIN_GO_VERSION} or later."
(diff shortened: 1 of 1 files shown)


Iceber:
The actual minimum version of clusterpedia is 1.17, the minimum version in the go mod is not updated, this is a new issue, you can fix it here or submit a new pr

> Caused by https://github.com/kubernetes-sigs/controller-runtime/blob/v0.12.1

ONE7live:
@Iceber Done, the minimum check version has been modified to 1.17

Iceber:
@ONE7live It seems that the go version check can be ensured by directly modifying the version in the go mod.
https://github.com/clusterpedia-io/clusterpedia/blob/478e2208efdeef96f69ca024392ad85f456b166f/go.mod#L1-L4

to
```
go 1.17
```

What do you think?

ONE7live:
@Iceber It may not be completely. In my personal opinion, we can indeed use go mod to ensure the go version used by clusterpedia, but we cannot guarantee that the go version of the user's environment matches it. For example, when the user executes update-vendor.sh to download dependencies, it may be more friendly to perform some environmental checks in advance and give prompts than to directly download and then generate errors,  what do you think?

Iceber:
I'm using go 1.17, I changed go mod to 1.18 and running go mod vendor gives me an error
```bash
$ make vendor
./hack/update-vendor.sh
running 'go mod tidy'
go mod tidy: go.mod file indicates go 1.18, but maximum supported version is 1.17
make: *** [vendor] Error 1
```
Does this do the same thing as the script?

ONE7live:
It can be understood that the purpose is the same, just a few more tips.
image
The script is compared against the go version of the user's environment through a predefined minimum version.

Iceber:
I think it might be better to restrict it in the go mod, so that maintenance costs can be reduced and the minimum version of go will need to be modified in one place in the future.

ONE7live:
@Iceber Ok, thanks for your reply, I agree with you. Do you need me to close this PR?

Iceber:
@ONE7live Although we use go mod to set the minimum version of go, the current minimum version in go.mod is wrong and should be 1.17. 

Would you like to submit pr to fix it?

PR #313, opened 2022-08-15
fix code-generator shell output flags

Icarus9913:
Signed-off-by: Icarus9913 <icaruswu66@qq.com>

**What type of PR is this?**

Add one of the following kinds:
/kind bug


Icarus9913:
https://github.com/kubernetes/code-generator/blob/97fa351e57d388212bfbf0b4249480b6db9ad79c/cmd/go-to-protobuf/protobuf/cmd.go#L91-L92

PR #323, opened 2022-08-18
move collection_resources from intelnalstorage to util

wuyingjun-lucky:

**What type of PR is this?**
/kind cleanup

Fixes #

```release-note
NONE
```


wuyingjun-lucky:
I think we can move `collection_resouce.go` to `util` that we can share `CollectionResources ` if we introduce other storage-layer to cluster-pedia.
I have not made a test yet. @Iceber If you think we can accept this change. I will test it on weekends on my side.

Iceber:
Currently, the storage layer determines which collection resources are supported.

 I feel that clusterpedia does not need to require which collection resources the storage layer must implement for the time being. 
The implementation of different storage layers can vary greatly, and it may be difficult to require them to implement certain collection resources

Issue #309, opened 2022-08-10
Apiserver address is null when pediacluster is applied by kubeconfig


wuyingjun-lucky:
### What would you like to be added?

Apiserver always be the real address.

### Why is this needed?

1、apiserver set to “”
<img width="627" alt="image" src="https://user-images.githubusercontent.com/16109961/183835433-408c9c21-e7e3-401a-8202-76ec9e3e04e9.png">

2、apiserver set to “dddd”
<img width="674" alt="image" src="https://user-images.githubusercontent.com/16109961/183835459-e3a5770b-a2f5-4b78-ad8d-4191693f7669.png">

wuyingjun-lucky:
@Iceber

Iceber:
Did you change the apiserver from the correct one to the wrong one?

> apiserver set to “dddd”

The conditoins and printer column were changed in v0.4.1, it will make the output look better, you can update the image version and crd yaml.

Also the string may be a domain name, maybe it doesn't work now, but it may work in the future, so we will allow the user to set any value, but if there is an exception, we will tell the user via conditons

> apiserver set to “”

The apiserver is allowed to be empty because the user can just provide kubeconfig. 
since we don't provide a multation webhook to parse the kubeconfig, the apiserver will be empty.
> Here are some discussions in https://github.com/clusterpedia-io/clusterpedia/pull/115#discussion_r828847480

wuyingjun-lucky:
> #115 (comment)

What I am confused about is why don`t we set the apiserver to the real address during reconcile cluster

Iceber:
> > #115 (comment)
> 
> What I am confused about is why don`t we set the apiserver to the real address during reconcile cluster

Extracting the apiserver address from kubeconfig seems like a good way to let the mutation webhook do the work. 
Try to only update the status when reconciling the cluster.

But we don't have a mutation webhook yet, so if it feels important to not have an `spec.apiserver` fields, we can check and set this field when reconciling the cluster.
> Add a mutation webhook for spec.apiserver, I'm not sure it's necessary

wuyingjun-lucky:
> But we don't have a mutation webhook yet, so if it feels important to not have an `spec.apiserver` fields, we can check and set this field when reconciling the cluster.
> 
> >

I think you are right, let webhook do the work is more appropriately.

May be we need a clusterpedia-ctl which can fix it easily.

Iceber:
[...]

PR #328, opened 2022-08-23
Add apiserver to status for display

wuyingjun-lucky:

**What type of PR is this?**
/kind feature

Fixes https://github.com/clusterpedia-io/clusterpedia/issues/309

```release-note
NONE
```

### DIFF
--- a/pkg/synchromanager/clustersynchro_manager.go
+++ b/pkg/synchromanager/clustersynchro_manager.go
@@ -269,7 +269,7 @@ func (manager *Manager) reconcileCluster(cluster *clusterv1alpha2.PediaCluster) 
 	config, err := buildClusterConfig(cluster)
 	if err != nil {
 		klog.ErrorS(err, "Failed to build cluster config", "cluster", cluster.Name)
-		manager.UpdateClusterValidatedCondition(cluster.Name, synchro, clusterv1alpha2.InvalidConfigReason,
+		manager.UpdateClusterAPIServerAndValidatedCondition(cluster.Name, cluster.Spec.APIServer, synchro, clusterv1alpha2.InvalidConfigReason,
 			"invalid cluster config: "+err.Error(), metav1.ConditionFalse)
 		return controller.NoRequeueResult
 	}
@@ -280,7 +280,7 @@ func (manager *Manager) reconcileCluster(cluster *clusterv1alpha2.PediaCluster) 
 		if ref, err := manager.clusterSyncResourcesLister.Get(refName); err != nil {
 			if !apierrors.IsNotFound(err) {
 				klog.ErrorS(err, "Failed to get SyncResourcesRef of cluster", "cluster", cluster.Name, "SyncResourcesRef", refName)
-				manager.UpdateClusterValidatedCondition(cluster.Name, synchro, clusterv1alpha2.InvalidSyncResourcesReason,
+				manager.UpdateClusterAPIServerAndValidatedCondition(cluster.Name, config.Host, synchro, clusterv1alpha2.InvalidSyncResourcesReason,
 					fmt.Sprintf("Failed to get cluster sync resources of cluster: %v", err), metav1.ConditionFalse)
 				return controller.RequeueResult(defaultRetryNum)
 			}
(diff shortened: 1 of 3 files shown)


pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -430,6 +430,9 @@ func (manager *Manager) UpdateClusterValidatedCondition(name string, synchro *cl
 
 func (manager *Manager) UpdateClusterStatus(ctx context.Context, name string, status *clusterv1alpha2.ClusterStatus) error {
 	return manager.updateClusterStatus(ctx, name, func(clusterStatus *clusterv1alpha2.ClusterStatus) {
+		if status.APIServer != "" {
+			clusterStatus.APIServer = status.APIServer
+		}
```
Iceber, line 435: `UpdateClusterValidatedCondition` is the first place to set the status, and it looks like status.apiserver needs to be set as well, otherwise if synchro fails to create, status.apiserver will always be empty.

wuyingjun-lucky: Is it ok to update `apiserver` in  `UpdateClusterValidatedCondition` ? 
`UpdateClusterValidatedCondition` seems  to be  be responsible for updating status condition.
And the `version`   also has not updated in `UpdateClusterValidatedCondition`.

Iceber: version is actually a runtime state (determined by the cluster), so it is better to set it in cluster synchro, we can also check the version once before creating cluster synchro but it is not necessary and will cause the apiserver of the member to be accessed every time it is reconciled.

The address of the apiserver is a static information that will not change as long as the apiserver or kubeconfig is not modified, and we can just set it in `UpdateClusterValidatedCondition`.

However, this also makes the semantics of the `UpdateClusterValidatedCondition` less accurate, so we can change the function name to `UpdateClusterAPIServerAndValidatedCondition` or something else.

wuyingjun-lucky: > The version is actually a runtime state (determined by the cluster), so it is better to set it in cluster synchro, we can also check the version once before creating cluster synchro but it is not necessary and will cause the apiserver of the member to be accessed every time it is reconciled.
> 
> The address of the apiserver is a static information that will not change as long as the apiserver or kubeconfig is not modified, and we can just set it in `UpdateClusterValidatedCondition`.
> 
> However, this also makes the semantics of the `UpdateClusterValidatedCondition` less accurate, so we can change the function name to `UpdateClusterAPIServerAndValidatedCondition` or something else.

OK， Do you think we should reserve the apiserver-updating in cluster synchro

Iceber: > OK， Do you think we should reserve the apiserver-updating in cluster synchro

It doesn't change in the cluster synchro, so I think it's okay not to update the apiserver.

staging/src/github.com/clusterpedia-io/api/cluster/v1alpha2/types.go
```diff
@@ -107,6 +107,10 @@ type ClusterGroupResources struct {
 }
 
 type ClusterStatus struct {
+	// +required
+	// +kubebuilder:validation:Required
+	APIServer string `json:"apiserver,omitempty"`
```
Iceber, line 112: The required field must not have an omitmepty tag, otherwise the update will fail if it equals ""

Iceber, line 112: The required field cannot have an omitmepty tag, otherwise it is not a required field in the generated CRD.
Also would it be better to allow apiserver to be empty, as an optional field?

The following Version should be an optional field, submit a new PR to fix it if you wish.

wuyingjun-lucky: updated 
or we can make a good-first-issue with `Version`

Iceber (commented):
Other looks good

staging/src/github.com/clusterpedia-io/api/cluster/v1alpha2/types.go
Iceber: good idea

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -269,7 +269,7 @@ func (manager *Manager) reconcileCluster(cluster *clusterv1alpha2.PediaCluster)
 	config, err := buildClusterConfig(cluster)
 	if err != nil {
 		klog.ErrorS(err, "Failed to build cluster config", "cluster", cluster.Name)
-		manager.UpdateClusterValidatedCondition(cluster.Name, synchro, clusterv1alpha2.InvalidConfigReason,
+		manager.UpdateClusterAPIServerAndValidatedCondition(cluster.Name, "", synchro, clusterv1alpha2.InvalidConfigReason,
```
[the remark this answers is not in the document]
wuyingjun-lucky: sounds good

PR #329, opened 2022-08-24
Remove Redundant Labels of Core Components

RuliXu:

**What type of PR is this?**
In `Helm` logic, what we do really care is `deployment`, `rs` does not matter that much.

Before:
image

After:
image

Fixes #

```release-note

```


Iceber:
Will service be affected when matching labels?

RuliXu:
> Will service be affected when matching labels?

Service will not be affected, as its selector only contains `app: clusterpedia-apiserver`.
image

PR #335, opened 2022-08-31
sync apiregistration resources

cleverhu:
**What type of PR is this?**
add apiregistration to discovery cache to support sync apiregistration resources

The kube-aggregator does not register `apiregistration.k8s.io` as an APIService resource, we need to add `apiregistration.k8s.io/v1` to the **groupVersions**

Fixes #

```release-note

```

### DIFF
--- a/pkg/synchromanager/clustersynchro/apiservice_controller.go
+++ b/pkg/synchromanager/clustersynchro/apiservice_controller.go
@@ -145,6 +145,9 @@ func (c *APIServiceController) reconcile() {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/synchromanager/clustersynchro/apiservice_controller.go
```diff
@@ -145,6 +145,9 @@ func (c *APIServiceController) reconcile() {
 		}
 	}
 
+	// add apiregistration to the group because apiregistrationinformers will not discover itself.
```
Iceber, line 148: suggest:
```go
// The kube-aggregator will not register `apiregistration.k8s.io` as an APIService resource, we need to add `apiregistration.k8s.io/v1` to the **groupVersions**
```

PR #350, opened 2022-09-14
support order by raw sql

cleverhu:
**What type of PR is this?**
support order by raw sql

The previous sorting can only be performed according to the specified fields, which limits the actual use.Therefore, it is necessary to support user-defined fields

Fixes #

```release-note
None
```

### DIFF
--- a/pkg/storage/internalstorage/util_test.go
+++ b/pkg/storage/internalstorage/util_test.go
@@ -374,7 +374,7 @@ func TestApplyListOptionsToQuery_OrderBy(t *testing.T) {
 		expected expected
 	}{
 		{
-			"asec",
+			"asc",
 			[]internal.OrderBy{
 				{Field: "namespace"},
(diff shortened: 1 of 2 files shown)


Iceber:
Can you add a few test cases for sorting fields in json?

pkg/storage/internalstorage/util_test.go
```diff
@@ -402,15 +402,24 @@ func TestApplyListOptionsToQuery_OrderBy(t *testing.T) {
 			},
 		},
 		{
-			"with not support fields",
+			"order by custom fields asc",
 			[]internal.OrderBy{
-				{Field: "name"},
-				{Field: "group"},
-				{Field: "cluster"},
```
Iceber, line 409: We can add a case where there are common fields and custom fields ordered together

Iceber (commented):
Other looks good to me

Issue #326, opened 2022-08-23
Support to create database when an existing storage component is used


27149chen:
### What would you like to be added?

Support to create database when an existing storage component is used

### Why is this needed?

clusterpedia may work together with other system which already has a storage component, to better share this component, clusterpedia should support to create batabase wth the permission of the user

Iceber:
We can add a flag to values, such as createDatabase, which defaults to false and needs to be actively set to true by the user to create the database, to ensure that the user is explicitly aware that a high privilege operation is being performed

`createDatabase=true` will render an init container for creating the database into the deployment.

PR #372, opened 2022-09-16
Support Create Database to External Storage

RuliXu:

**What type of PR is this?**
/kind feature

Clusterpedia may work together with other system which already has a storage component, to better share this component, clusterpedia should support to create database with the permission of the user.

This PR adds a flag `createDatabase` to values, which defaults to be false and needs to be actively set to true by the user to create the database, to ensure that the user is explicitly aware that a high privilege operation is being performed.

`--set externalStorage.createDatabase=true` will render an init container for creating the database into the core component deployments.

https://github.com/clusterpedia-io/clusterpedia/issues/326

### Mysql
image

### Postgresql
image

### Mysql
image

### Postgresql
image

```release-note

```

### DIFF
--- a/charts/clusterpedia/templates/_helpers.tpl
+++ b/charts/clusterpedia/templates/_helpers.tpl
@@ -188,7 +188,7 @@ Return the proper Docker Image Registry Secret Names
      {{- if empty .Values.externalStorage.database }}
           {{ required "Please set correct storage database!" "" }}
      {{- else -}}
-          {{- .Values.externalStorage.database | quote -}}
+          {{- .Values.externalStorage.database }}
      {{- end -}}
 {{- else -}}
      {{- "clusterpedia" -}}
@@ -304,5 +304,13 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
      {{- .Values.postgresql.primary.persistence.selector.matchLables -}}
 {{- else if eq (include "clusterpedia.storage.type" .) "mysql" -}}
      {{- .Values.mysql.primary.persistence.selector.matchLabels -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "clusterpedia.storage.password.envKey" -}}
+{{- if eq (include "clusterpedia.storage.type" .) "postgres" }}
+     {{- "PGPASSWORD" -}}
+{{- else if eq (include "clusterpedia.storage.type" .) "mysql" -}}
+     {{- "DB_PASSWORD" -}}
 {{- end -}}
 {{- end -}}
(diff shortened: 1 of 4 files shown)


charts/clusterpedia/templates/_helpers.tpl
```diff
@@ -306,3 +306,11 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
      {{- .Values.mysql.primary.persistence.selector.matchLabels -}}
 {{- end -}}
 {{- end -}}
+
+{{- define "clusterpedia.storage.password.name" -}}
```
Iceber, line 310: Is `clusterpedia.storage.envKeys.password` or `clusterpedia.storage.envs.password` better?

charts/clusterpedia/templates/apiserver-deployment.yaml
```diff
@@ -30,6 +30,32 @@ spec:
...
+            echo 'DataBase OK ✓'
+          {{- end }}
+        env:
+          - name: {{ include "clusterpedia.storage.password.name" . }}
+            valueFrom:
+              secretKeyRef:
+                name: {{ include "clusterpedia.internalstorage.fullname" . }}
+                key: password
+    {{- end }}
       - name: check-storage
```
Iceber, line 59: `ensure-database` seems to have some overlap with `check-storage`. 
Could we not run the init container `check-storage` when the user needs to create a database?

Iceber: Or the init container `check-storage` is run first and only checks if mysql/postgres is ready,

The pg_isready connects to the postgres database, not to the user-specified database
>           - until pg_isready -U {{ include "clusterpedia.storage.user" . }} -d "dbname={{ include "clusterpedia.storage.database" . }}" -h {{ include "clusterpedia.storage.host" . }} -p {{ include "clusterpedia.storage.port" . }}; do sleep 1; done

charts/clusterpedia/templates/_helpers.tpl
RuliXu: Indeed, maybe `clusterpedia.storage.password.envKeys` is better

charts/clusterpedia/templates/apiserver-deployment.yaml
RuliXu: Maybe we can run only 1 container, if `ensure-database` is rendered, `check-storage` can be replaced.

charts/clusterpedia/templates/_helpers.tpl
Iceber: This is good, but `envKey` should be singular.

Iceber:
/lgtm
@27149chen PTAL, Thanks

PR #413, opened 2022-10-10
Add Logger Config To Charts

RuliXu:

**What type of PR is this?**

Fixes # https://github.com/clusterpedia-io/clusterpedia/issues/13

image
image

```release-note
This PR changes default MaxBackups from 1 to 0 in the code to stay the same with the default lumberjack logger behavior.
```

### DIFF
--- a/charts/clusterpedia/values.yaml
+++ b/charts/clusterpedia/values.yaml
@@ -58,6 +58,16 @@ storageConfig:
     slowThreshold: 100ms
     ## @param storageConfig.log.ignoreRecordNotFoundError indicates whether ignore error type `NotFound`
     ignoreRecordNotFoundError: false
+    ## @param storageConfig.log.logger indicates the log rotate config
+    logger:
+      ## @param storageConfig.log.logger.filename indicates the file to write logs to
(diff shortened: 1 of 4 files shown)


charts/clusterpedia/values.yaml
```diff
@@ -58,6 +58,16 @@ storageConfig:
...
+    ## @param storageConfig.log.logger indicates the log rotate config
+    logger:
+      ## @param storageConfig.log.logger.filename indicates the file to write logs to
+      filename: /var/log/clusterpedia/internalstorage.log
+      ## @param storageConfig.log.logger.maxSize indicates the maximum size in megabytes of the log file before it gets
+      ## rotated
+      maxSize: 100
+      ## @param storageConfig.log.logger.maxBackups indicates the maximum number of old log files to retain. Default 0
+      ## is to retain all old log files
+      maxBackups: 0
```
Iceber, line 70: The default value of 1 in the code, what do you think is better

RuliXu: > The default value of 1 in the code, what do you think is better

Thanks for reminding, let's keep the same with the default lumberjack logger behavior, which is `0`.

Issue #397, opened 2022-09-28
An unexpected error occured when getting resources with a fake cluster name


27149chen:
### What happened?

when getting resources with a fake cluster name (which does not exist in clusterpedia) using clusterclient
The following error is returned:
no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1"

### What did you expect to happen?

return a "cluster not found" error

### How can we reproduce it (as minimally and precisely as possible)?

getting resources with a fake cluster name (which does not exist in clusterpedia) using clusterclient

### Clusterpedia Version

<details>

```console
# replace the pod name of clusterpedia apiserver
$ kubectl -n clusterpedia-system exec {clusterpedia-apiserver-pod-name} -- apiserver --version=raw
# paste output here

# replace the pod name of clusterpedia clustersynchro-manager
$ kubectl -n clusterpedia-system exec {clustersynchro-manager-pod-name} -- clustersynchro-manager --version=raw
# paste output here
```

</details>

### Host Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

qiuming520:
@27149chen 
```no matches for kind "CustomResourceDefinition"``` no matches for kind "CustomResourceDefinition" that implies that the version of kubernetes you're using doesn't have the CRD type available. you need to  confirm what version or flavor kube you have installed

27149chen:
@qiuming520  Sorry, I find the problem is not in "collection resources", but in the clustered client, if I new a client using "ClusterConfigFor" with an invalid cluster name, and then use this client to list resources (for example, list crds), the error above is raised

Iceber:
We do not currently return a specific error when a cluster does not exist.

It seems to make sense to tell the user that the requested cluster does not exist

PR #431, opened 2022-10-25
check cluster whether exist

lengrongfu:
**What type of PR is this?**
/kind bug

When Cluster does not exist, return an error inaccurate.

Fixes https://github.com/clusterpedia-io/clusterpedia/issues/397

```release-note

```

### DIFF
--- a/pkg/kubeapiserver/resource_handler.go
+++ b/pkg/kubeapiserver/resource_handler.go
@@ -49,6 +49,29 @@ func (r *ResourceHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	gvr := schema.GroupVersionResource{Group: requestInfo.APIGroup, Version: requestInfo.APIVersion, Resource: requestInfo.Resource}
 
 	clusterName := request.ClusterNameValue(req.Context())
+
+	var (
+		cluster *clusterv1alpha2.PediaCluster
+		err     error
+	)
+	// When clusterName not empty, first check cluster whether exist
+	if clusterName != "" {
+		if cluster, err = r.clusterLister.Get(clusterName); err != nil {
+			if !apierrors.IsNotFound(err) {
+				klog.ErrorS(err, "Failed to handle resource request, not get cluster from cache", "cluster", clusterName, "resource", gvr)
+				responsewriters.ErrorNegotiated(
+					apierrors.NewInternalError(err),
+					Codecs, gvr.GroupVersion(), w, req,
+				)
+				return
+			}
+			responsewriters.ErrorNegotiated(
+				apierrors.NewBadRequest("not found requested cluster"),
+				Codecs, gvr.GroupVersion(), w, req,
+			)
+			return
+		}
+	}
 	if !r.discovery.ResourceEnabled(clusterName, gvr) {
 		r.delegate.ServeHTTP(w, req)
 		return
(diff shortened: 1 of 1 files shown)


pkg/kubeapiserver/resource_handler.go
```diff
@@ -73,17 +86,7 @@ func (r *ResourceHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
-			err := fmt.Errorf("not found request cluster")
-			klog.ErrorS(err, "Failed to handle resource request, not get cluster from cache", "cluster", clusterName, "resource", gvr)
-			responsewriters.ErrorNegotiated(
-				apierrors.NewInternalError(err),
-				Codecs, gvr.GroupVersion(), w, req,
-			)
-			return
-		}
-
+		cluster, _ := r.clusterLister.Get(clusterName)
```
Iceber, line 89: It is possible to use the cluster got in the previous step, because we can't guarantee that getting it from Lister again will return the cluster

pkg/kubeapiserver/resource_handler.go
```diff
@@ -49,6 +49,19 @@ func (r *ResourceHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	gvr := schema.GroupVersionResource{Group: requestInfo.APIGroup, Version: requestInfo.APIVersion, Resource: requestInfo.Resource}
 
 	clusterName := request.ClusterNameValue(req.Context())
+	// When clusterName not empty, first check cluster whether exist
+	if clusterName != "" {
+		if _, err := r.clusterLister.Get(clusterName); err != nil {
+			klog.ErrorS(err, "Failed to handle resource request, not get cluster from cache", "cluster", clusterName, "resource", gvr)
+			err = fmt.Errorf("not found request cluster")
+			responsewriters.ErrorNegotiated(
+				apierrors.NewInternalError(err),
```
Iceber, line 58: If the error is cluster not found, then instead of returning InternalError, we return BadRequest(NewBadRequest) and do not need to print the log.

pkg/kubeapiserver/resource_handler.go
lengrongfu: done

pkg/kubeapiserver/resource_handler.go
lengrongfu: done

pkg/kubeapiserver/resource_handler.go
```diff
@@ -73,17 +96,6 @@ func (r *ResourceHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 
 	// Check the health of the cluster
 	if clusterName != "" {
```
Iceber, line 98: Since the `cluster` will be used, it might be better to use `cluster ! = nil`

pkg/kubeapiserver/resource_handler.go
```diff
@@ -49,6 +49,29 @@ func (r *ResourceHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+			if !apierrors.IsNotFound(err) {
+				klog.ErrorS(err, "Failed to handle resource request, not get cluster from cache", "cluster", clusterName, "resource", gvr)
+				responsewriters.ErrorNegotiated(
+					apierrors.NewInternalError(err),
+					Codecs, gvr.GroupVersion(), w, req,
+				)
+				return
+			}
+			responsewriters.ErrorNegotiated(
+				apierrors.NewBadRequest(err.Error()),
```
Iceber, line 69: For errors that have been identified, it is better to use a fixed format or string, for example `not found requested cluster`

Iceber:
/lgtm

Thanks for your contribution

PR #445, opened 2022-11-07
storage: change the Watch method to use *internal.ListOptions

Iceber:

**What type of PR is this?**

/kind feature
`internal.ListOptions` has more information than `metainternalversion.ListOptions`
Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/storage/internalstorage/resource_storage.go
+++ b/pkg/storage/internalstorage/resource_storage.go
@@ -9,8 +9,8 @@ import (
 	"strconv"
 
 	"gorm.io/gorm"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
-	metainternalversion "k8s.io/apimachinery/pkg/apis/meta/internalversion"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/conversion"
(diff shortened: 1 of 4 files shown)


pkg/storage/internalstorage/resource_storage.go
```diff
@@ -274,8 +274,8 @@ func (s *ResourceStorage) List(ctx context.Context, listObject runtime.Object, o
 	return nil
 }
 
-func (s *ResourceStorage) Watch(ctx context.Context, options *metainternalversion.ListOptions) (watch.Interface, error) {
-	return nil, nil
+func (s *ResourceStorage) Watch(_ context.Context, _ *internal.ListOptions) (watch.Interface, error) {
+	return nil, apierrors.NewMethodNotSupported(s.storageGroupResource, "watch")
```
wuyingjun-lucky, line 278: By the way 
Do you have any idea about watch in internal- storage-layer ？

pkg/storage/memorystorage/memory_resource_storage.go
```diff
@@ -121,7 +120,7 @@ func (s *ResourceStorage) Get(ctx context.Context, cluster, namespace, name stri
 	return nil
 }
 
-//nolint
+// nolint
```
wuyingjun-lucky, line 123: We can add the nolint rule (specific linter) next pr.  @xyz2277

pkg/storage/internalstorage/resource_storage.go
Iceber: This will be something we need to discuss and advance recently

My initial idea is to decouple the resource events between apiserver and synchronizer with the help of some message queue components

wuyingjun-lucky:
I want add e2e test  for list & watch.. in deferent situation as soon as possible. 
@Iceber   Do you think is it necessary. 
@qiuming520

pkg/storage/internalstorage/resource_storage.go
wuyingjun-lucky: > This will be something we need to discuss and advance recently
> 
> My initial idea is to decouple the resource events between apiserver and synchronizer with the help of some message queue components

We will enlarge member clusters scale next years. I think we may change the storage in future.

Iceber: I want to separate internalstorage from the clusterpedia repo so that we can iterate on internalstorage separately and deliver new features faster

What do you think?

Iceber:
> I want add e2e test for list & watch.. in deferent situation as soon as possible.
@Iceber Do you think is it necessary.

I think it's great that whenever more tests are always good.

hanweisen:
LGTM

PR #460, opened 2022-11-17
Readiness probe with standalone tcp

Iceber:
**What type of PR is this?**

/kind feature
When client-go creates any number of Clients with the same configuration, such as certificates, it reuses the same TCP connection.
https://github.com/kubernetes/kubernetes/blob/3f823c0daa002158b12bfb2d53bcfe433516659d/staging/src/k8s.io/client-go/transport/transport.go#L54

This results in the cluster health check interface using the same TCP connection as the resource synchronized informer, which may cause TCP blocking and increased health check latency if a large number of informers are started for the first time.
We add a feature gate - `ClusterReadinessProbeWithStandaloneTCP` to allow users to use a standalone tcp for health checks

```bash
./clustersynchro-manager --feature-gates=ClusterReadinessProbeWithStandaloneTCP=true
```

Fixes #

```release-note
NONE
```


Iceber:
@cleverhu PTAL

PR #465, opened 2022-11-29
make synchromanager worker configurable 

### LINKED ISSUE #458: make synchromanager worker configurable

1. Add option WorkerNumber  for `clustersynchro-manager` such as 
```
WorkerNumber int          // WorkerNumber is the number of worker goroutines
```
2. change the code from `synchromanager.Run(1, ctx.Done())` to `synchromanager.Run(c.WorkerNumber, ctx.Done())` in 
`pkg/synchromanager/clustersynchro_manager.go`
3. If the WorkerNumber is less than or equal to 0, we should set the default value to 3.

### Why is this needed?

Now  the worker goroutine number for the cluster synchronization controller in clustersynchro manager is hardcode to 1, which results in the cluster init will become slowly if the user joined too many clusters. Therefore, the worker goroutine number should be configurable.

wlp1153468871:
**What type of PR is this?**

/kind feature

Fixes #458

```release-note
make synchromanager worker configurable 
```

### DIFF
--- a/cmd/clustersynchro-manager/app/options/options.go
+++ b/cmd/clustersynchro-manager/app/options/options.go
@@ -1,6 +1,8 @@
 package options
 
 import (
+	"fmt"
+
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -32,8 +34,9 @@ type Options struct {
 	LeaderElection   componentbaseconfig.LeaderElectionConfiguration
 	ClientConnection componentbaseconfig.ClientConnectionConfiguration
 
-	Logs    *logs.Options
-	Storage *storageoptions.StorageOptions
+	Logs         *logs.Options
+	Storage      *storageoptions.StorageOptions
+	WorkerNumber int // WorkerNumber is the number of worker goroutines
 
 	Master     string
 	Kubeconfig string
@@ -65,6 +68,7 @@ func NewClusterSynchroManagerOptions() (*Options, error) {
 
 	options.Logs = logs.NewOptions()
 	options.Storage = storageoptions.NewStorageOptions()
+	options.WorkerNumber = 5
 	return &options, nil
 }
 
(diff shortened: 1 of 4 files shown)


cmd/clustersynchro-manager/app/options/options.go
```diff
@@ -65,6 +68,7 @@ func NewClusterSynchroManagerOptions() (*Options, error) {
 
 	options.Logs = logs.NewOptions()
 	options.Storage = storageoptions.NewStorageOptions()
+	options.WorkerNumber = 1
```
Iceber, line 71: The default number of workers is 1 which is not really suitable, we need to set a larger default value, for example 5 or 15.
@cleverhu @wlp1153468871 what do you think?

cmd/clustersynchro-manager/app/options/options.go
```diff
@@ -75,6 +79,7 @@ func (o *Options) Flags() cliflag.NamedFlagSets {
 	genericfs.StringVar(&o.ClientConnection.ContentType, "kube-api-content-type", o.ClientConnection.ContentType, "Content type of requests sent to apiserver.")
 	genericfs.Float32Var(&o.ClientConnection.QPS, "kube-api-qps", o.ClientConnection.QPS, "QPS to use while talking with kubernetes apiserver.")
 	genericfs.Int32Var(&o.ClientConnection.Burst, "kube-api-burst", o.ClientConnection.Burst, "Burst to use while talking with kubernetes apiserver.")
+	genericfs.IntVar(&o.WorkerNumber, "worker-number", o.WorkerNumber, "The number of worker goroutines")
```
cleverhu, line 82: It might be better to end with `.`.

cmd/clustersynchro-manager/app/options/options.go
cleverhu: I think it should be greater than 1 such as 3 or 5.

wlp1153468871:
updata

Iceber:
/lgtm
@wlp1153468871 Thanks for your contribution!

PR #478, opened 2022-12-30
use errors.Is to check for a specific error

Fish-pro:
Signed-off-by: Fish-pro <zechun.chen@daocloud.io>

**What type of PR is this?**

/kind cleanup

Comparing with `==` will fail on wrapped errors. Use errors.Is to check for a specific error

Fixes #

```release-note
NONE
```


Iceber:
We can't modify the reflector.go alone, we need to keep it as consistent as possible with client-go's reflector.go, so if we need to modify it, we can create pr in `client-go` and then clusterpedia will synchronize it with client-go's dependencies

Fish-pro:
> We can't modify the reflector.go alone, we need to keep it as consistent as possible with client-go's reflector.go, so if we need to modify it, we can create pr in `client-go` and then clusterpedia will synchronize it with client-go's dependencies

@Iceber Thanks for your reply, I got it

PR #479, opened 2022-12-30
add retry for mysql read-only error

cleverhu:
**What type of PR is this?**

/kind bug

Add retry for mysql read-only error. Therefore, if the database is read-only for some reason, it will attempt to recover for a while.

Fixes #

```release-note

```

### DIFF
--- a/pkg/storage/internalstorage/errors.go
+++ b/pkg/storage/internalstorage/errors.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net"
 	"os"
+	"sync"
 	"syscall"
 
 	"github.com/go-sql-driver/mysql"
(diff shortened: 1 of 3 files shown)


pkg/storage/internalstorage/errors.go
```diff
@@ -18,6 +19,11 @@ import (
 	"github.com/clusterpedia-io/clusterpedia/pkg/storage"
 )
 
+var (
+	recoverableMysqlErrNumbers  sync.Map
+	recoverablePostgresErrCodes sync.Map
+)
+
```
Iceber, line 26: It is recommended that the default recoverable error be maintained uniformly
```go
func init() {
        recoverableMysqlErrNumbers.Store(1053, struct{}{}) // ER_SERVER_SHUTDOWN: Server shutdown in progress
        recoverableMysqlErrNumbers.Store(1205, struct{}{}) // Error 1205: Lock wait timeout exceeded; try restarting transaction.
        recoverableMysqlErrNumbers.Store(1290, struct{}{}) // Error 1290: The MySQL server is running with the --read-only option so it cannot execute this statement.

        recoverablePostgresErrCodes.Store(pgerrcode.AdminShutdown, struct{}{})
}
```

Iceber (commented):
others lgtm

PR #495, opened 2023-02-14
optimize calculate paging count

cleverhu:
**What type of PR is this?**

/kind feature

Calculate the paging length from select count (` object ') to select count (*) to speed-up response.

Before:
83FB02FC824F81A02F03CC056ACDE05B
After:
企业微信截图_4fd72047-3130-42f9-9971-35d4ec44e0dc

Fixes #

```release-note

```

### DIFF
--- a/pkg/storage/internalstorage/collectionresource_storage.go
+++ b/pkg/storage/internalstorage/collectionresource_storage.go
@@ -66,7 +66,7 @@ func (s *CollectionResourceStorage) query(ctx context.Context, opts *internal.Li
 
 	query := s.db.WithContext(ctx).Model(&Resource{})
 	if s.typesQuery != nil {
-		return result.Select(query).Where(s.typesQuery), result, nil
+		return query.Where(s.typesQuery), result, nil
 	}
 
 	// The `URLQueryGroups` and `URLQueryResources` only works on *Any Collection Resource*,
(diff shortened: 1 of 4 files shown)


pkg/storage/internalstorage/util.go
```diff
@@ -158,7 +158,13 @@ func applyListOptionsToQuery(query *gorm.DB, opts *internal.ListOptions, applyFn
...
 	if opts.WithRemainingCount != nil && *opts.WithRemainingCount {
 		amount = new(int64)
-		query = query.Count(amount)
+		countQuery := *query
+		statement := *countQuery.Statement
+		countQuery.Statement = &statement
+		if len(countQuery.Statement.Selects) == 1 && countQuery.Statement.Selects[0] == objectColumn {
+			countQuery.Statement.Selects = []string{"*"}
+		}
+		countQuery.Count(amount)
```
Iceber, line 167: This may seem like a very hack way of doing things, but in reality we want to skip the field check when `Count` and just put the select at the end.

https://github.com/clusterpedia-io/clusterpedia/blob/1d4658d1c4f3c2f815d335f7a79a3d7d5f39f3e5/pkg/storage/internalstorage/resource_storage.go#L199-L200
->
```go
        if err := result.From(result.Select(query)); err != nil {
                return InterpretDBError(s.storageGroupResource.String(), err)
        }
```

Iceber (changes requested):

Iceber (approved):
LGTM

PR #541, opened 2023-07-14
:seedling: add unit tests for events.go

khareyash05:
**What type of PR is this?**
/kind feature

Adds tests for `pkg/synchromanager/clustersynchro/queue/event.go`

```release-note
NONE
```

### DIFF
--- a/pkg/synchromanager/clustersynchro/queue/event_test.go
+++ b/pkg/synchromanager/clustersynchro/queue/event_test.go
@@ -8,6 +8,7 @@ type pressureEventsTest struct {
 	name       string
 	older      *Event
 	newer      *Event
+	expected   *Event
 	reputCount int
 }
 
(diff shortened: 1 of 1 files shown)


pkg/synchromanager/clustersynchro/queue/event_test.go
```diff
@@ -0,0 +1,137 @@
...
+				reputCount: 1,
+				Action:     Added,
+				Object:     "Object1",
+			},
+			newer: &Event{
+				reputCount: 2,
+				Action:     Updated,
+				Object:     "Object2",
+			},
+			reputCount: 2,
```
Iceber, line 28: We might be better off with the full event as expected, even though the data and reputCount are the same as the newer, but the Action might change, for example here the event Action would be `Added`

khareyash05: I am not sure if I understood it correctly.Do I have to add more tests as well?

Iceber: I think we can use an expected event object to compare, e.g.
```go
			name: "Newer event is not nil",
			older: &Event{
				reputCount: 1,
				Action:     Added,
				Object:     "Object1",
			},
			newer: &Event{
				reputCount: 2,
				Action:     Updated,
				Object:     "Object2",
			},
			expected: Event{
				reputCount: 2,
				Action:     Added,
				Object:     "Object2",
			},
````

Iceber (approved):
LGTM, Thank you very much for contributing.

Iceber:
@khareyash05 golang-lint is falied.

khareyash05:
Fixed Lint.

PR #551, opened 2023-07-28
:seedling:  upgrade to Kustomize configuration

khareyash05:
**What type of PR is this?**
/kind feature

To switch from Helm Charts to Kustomize

```release-note
NONE
```


Iceber:
It is best to keep the contents of the deploy directory and the kustomize file in */kustomize* to ensure that */deploy* can still be deployed with the old documentation when kustomize is not fully available

khareyash05:
> It is best to keep the contents of the _/deploy_ directory and the kustomize file in _/kustomize_ to ensure that _/deploy_ can still be deployed with the old documentation when kustomize is not fully available

will do that

kustomize/binding-apiserver/binding-apiserver.yaml
```diff
@@ -0,0 +1,88 @@
+apiVersion: v1
```
Iceber, line 1: We shouldn't have to keep the yaml file for the build

kustomize/clusterpedia_apiserver.yaml
```diff
@@ -0,0 +1,73 @@
+apiVersion: v1
```
Iceber, line 1: They can move into */kustomize/apiserver*

kustomize/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- ./binding-apiserver
+- ./apiserver
+- ./clustersynchro-manager
+- ./controller-manager
+- ./crds
+- ./internalstorage
```
Iceber, line 10: We can only choose one storage component when deploying, so we can't specify internalstorage here, but rather internalstorage/mysql or internalstorage/postgres

kustomize/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- ./binding-apiserver
+- ./apiserver
```
Iceber, line 6: apiserver and binding-apiserver cannot be deployed at the same time, we don't need to install `binding-apiserver` here.

kustomize/internalstorage/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
Iceber, line 1: internalstorage doesn't need kustomize.yaml, we will only choose one of mysql or postgres in the installation.

Iceber (commented):
The others look great. :thumbsup:



<br class="Apple-interchange-newline">

kustomize/setup.yaml
```diff
@@ -0,0 +1,1110 @@
+apiVersion: v1
```
Iceber, line 1: We don't need this file generated by `kubectl kustomize`.

kustomize/apiserver/clusterpedia_apiserver.yaml
```diff
@@ -0,0 +1,73 @@
+apiVersion: v1
```
Iceber, line 1: Last issue, this file is also generated by `kubectl kustomie`

khareyash05: Missed that . will update

Iceber (approved):
Thank you very much for your contribution

PR #581, opened 2023-10-24
:seedling: migrate golang from 1.20 to 1.21.3

khareyash05:
**What type of PR is this?**
/kind deprecation

Bump Golang from 1.19 to 1.21 as 1.19 has reached end of life 
Fixes #NA

```release-note
NONE
```

### DIFF
--- a/builder.dockerfile
+++ b/builder.dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.20.7
+FROM golang:1.21.3
 
 # FROM golang:1.19.2-alpine
 # RUN apk add --no-cache build-base git bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v4
         with:
-          go-version: '1.20'
+          go-version: '1.21'
       - uses: golangci/golangci-lint-action@v3
         with:
           version: v1.53.3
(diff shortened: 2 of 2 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/clusterpedia-io/clusterpedia
 
-go 1.19
+go 1.21
```
Iceber, line 3: The version in go.mod represents the minimum version number, so I don't think we need to change it if we don't have to.

Of course we would normally update it based on the k8s dependencies

khareyash05: Oh but golang 1.19 isn't supported now, Can we shift it to 1.20? As minimum supported version

Iceber: I don't think it's necessary, the minimum version is more of a minimum version dependency in terms of functionality rather than a recommended version for users. We can use the correct version at build and test time

builder.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.20.7
+FROM golang:1.21.0
```
Iceber, line 1: We need to use the latest version `1.21.3`

go.mod
khareyash05: Understood

Iceber (approved):
lgtm， Thanks

Iceber:
@khareyash05 Looks like the commit info and pr title need to be updated

Iceber:
This pr is bump golang from 1.20 to 1.21.3, not 1.19

and the commit message isn't updated

khareyash05:
> This pr is bump golang from 1.20 to 1.21.3, not 1.19
> 
> and the commit message isn't updated

Shall I merge all in one commit and then push? Also came you suggest me a commit message

khareyash05:
@Iceber Updated the PR

Iceber:
> Shall I merge all in one commit and then push? Also could you suggest me a commit message

We try to use a single commit, unless splitting it makes it easier to understand.

Issue #589, opened 2023-11-10
Expose  binding-apiserver directly to end uses


huiwq1990:
### What would you like to be added?

1）Export binding-apiserver as apiserver directly to user, the server path not need add lusterpedia.io/v1beta1/resources suffix；
2）clusterpedia-apiserver support service nodeport;

In a word, we want to use binding-apiserver like this:

```shell

apiVersion: v1
kind: Config
clusters:
- name: default-cluster
  cluster:
    insecure-skip-tls-verify: true
    server: https://${binding-apiserve-ip}:${binding-apiserve-port}
contexts:
- name: default-context
  context:
    cluster: default-cluster
    namespace: default
    user: default-user
current-context: default-context
users:
- name: default-user
  user:
    token:  abc
```

The `binding-apiserve-ip` could be `clusterpedia-apiserver`  service ip or host cluster node ip.
The `binding-apiserve-port` could be `clusterpedia-apiserver`  service port or nodeport.

### Why is this needed?

In normal case, binding-apiserver works as a aggregated apiserver, and exposed by host apiserver. So all binding-apiserver requests must forward by host apiserver, and it effects host apiserver performance. In other case, we want to watch binding-apiserver, but after a while, we get some error.
```

I1109 09:23:51.743143       1 httplog.go:131] "HTTP" verb="GET" URI="/apis/clusterpedia.io/v1beta1/resources/api/v1/nodes?limit=500" latency="7.3116ms" userAgent="kubectl/v1.16.15 (linux/amd64) kubernetes/2adc8d7" audit-ID="4132ae59-b216-485f-8a83-2caa0d5c6085" srcIP="10.0.0.4:47654" apf_pl="workload-low" apf_fs="service-accounts" apf_iseats=1 apf_fseats=0 apf_additionalLatency="0s" apf_execution_time="7.111772ms" resp=200
[...]

PR #590, opened 2023-11-13
support apiserver url rewrite

huiwq1990:
**What type of PR is this?**

/kind feature

Expose the binding-apiserver more kubernete.

Fixes https://github.com/clusterpedia-io/clusterpedia/issues/589

1) Add command ` --feature-gates=ApiServerURLRewrite=true` to open the feature.
2) kubeconfig file's server: https://${binding-apiserver}:${binding-apiserver-port}/resources, example: https://10.0.0.4:30443/resources

```release-note

```


Iceber:
If `/resources` is still needed, I think it's not necessary to change it, and `/apis/clusterpedia.io/v1beta1/resources` is more standardized and easier to understand than `/resources`.

If you choose to remove the prefix, then `clusterpedia-ip:port/apis/<group>/<resource> `-> `clusterpedia-ip:port/apis/clusterpedia.io/v1beta1/resources/apis/<group>/<resource>`would be better.

huiwq1990:
> If `/resources` is still needed, I think it's not necessary to change it, and `/apis/clusterpedia.io/v1beta1/resources` is more standardized and easier to understand than `/resources`.
> 
> If you choose to remove the prefix, then `clusterpedia-ip:port/apis/<group>/<resource> `-> `clusterpedia-ip:port/apis/clusterpedia.io/v1beta1/resources/apis/<group>/<resource>`would be better.

@Iceber I made some change, and remove `/resources` suffix.

pkg/utils/filters/rewrite.go
```diff
@@ -0,0 +1,157 @@
...
+
+	return h
+}
+
+const OldResourceApiServerPrefix = "/apis/clusterpedia.io/v1beta1/resources"
+
+func WithRewriteFilter(handler http.Handler) http.Handler {
+	rh := NewHandler(map[string]string{
+		"/(.*)": fmt.Sprintf("%s/$1", OldResourceApiServerPrefix),
+	})
```
wzshiming, line 129: I think we can define a singleton constant for this

pkg/utils/filters/rewrite.go
```diff
@@ -0,0 +1,157 @@
...
+
+const OriginURIHeaderKey = "X-Rewrite-Original-URI"
+
+type Rule struct {
+	Pattern string
+	To      string
+	*regexp.Regexp
+}
+
+var regfmt = regexp.MustCompile(`:[^/#?()\.\\]+`)
```
wzshiming, line 22: I think we can replace the RE with a matching prefix to improve performance

pkg/utils/filters/rewrite.go
```diff
@@ -0,0 +1,157 @@
...
+		rh.ServeHTTP(w, req)
+		handler.ServeHTTP(w, req)
+	})
+}
+
+type RewriteHandler struct {
+	rules []*Rule
+}
+
+func (h *RewriteHandler) ServeHTTP(res http.ResponseWriter, req *http.Request) {
```
wzshiming, line 141:
```suggestion
func (h *RewriteHandler) Rewrite(req *http.Request) {
```

pkg/utils/filters/rewrite.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func hit(pattern, str string) bool {
+	r, e := regexp.MatchString(pattern, str)
+	if e != nil {
+		return false
+	}
+
+	return r
+}
```
wzshiming, line 157: The pattern here will be compiled multiple times, we should avoid it.

huiwq1990:
@wzshiming The old version, I want to implement a general rewriter, it's too complex for us. So I did some refactor, it's easy to understand.

pkg/utils/filters/rewrite_test.go
```diff
@@ -0,0 +1,77 @@
...
+	urls []kubeRequest
+}
+
+type kubeRequest struct {
+	from string
+	to   string
+}
+
+func TestRewrite(t *testing.T) {
+	tests := []testCase{
```
wzshiming, line 20: Add a case
```
urls: []kubeRequest{
  {
    from: "/apis/clusterpedia.io/v1beta1/resourcesany",
    to: "/apis/clusterpedia.io/v1beta1/resources/apis/clusterpedia.io/v1beta1/resourcesany",
  },
  {
    from: "/apis/clusterpedia.io/v1beta1/resources",
    to: "/apis/clusterpedia.io/v1beta1/resources",
  },
},
```

pkg/utils/filters/rewrite.go
```diff
@@ -0,0 +1,40 @@
...
+	if err != nil {
+		return false
+	}
+	req.URL.Path = rewritePath
+	req.Header.Set(OriginPathHeaderKey, oldPath)
+
+	klog.V(5).InfoS("Rewrite url", "oldPath", oldPath, "newPath", req.URL.EscapedPath())
+
+	return true
+}
```
wzshiming, line 40: I thinks we just need to cover this function in unit tests without starting the http server

```suggestion
func urlPrefixRewrite(oldPath string) (string, bool) {
	if strings.HasPrefix(oldPath, OldResourceApiServerPrefix) {
		return "", false
	}

	rewritePath, err := url.JoinPath(OldResourceApiServerPrefix, req.URL.Path)
	if err != nil {
		return "", false
	}

	return rewritePath, true
}
```

pkg/utils/filters/rewrite.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+const OriginPathHeaderKey = "X-Rewrite-Original-Path"
+const OldResourceApiServerPrefix = "/apis/clusterpedia.io/v1beta1/resources"
+
+func WithRewriteFilter(handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		if !doUrlRewrite(req) {
+			klog.V(5).InfoS("request not need rewrite", "path", req.URL.EscapedPath())
+		}
```
wzshiming, line 18:
```suggestion
		oldPath := req.URL.EscapedPath()
		if newPath, ok := urlPrefixRewrite(req); ok {
			req.URL.Path = rewritePath
			req.Header.Set(OriginPathHeaderKey, oldPath)
			klog.V(5).InfoS("Rewrite url", "oldPath", oldPath, "newPath", req.URL.EscapedPath())
		} else {
			klog.V(5).InfoS("request not need rewrite", "path", oldPath)
		}
```

wzshiming:
LGTM

/cc @Iceber

pkg/apiserver/features/features.go
```diff
@@ -0,0 +1,26 @@
...
+	"k8s.io/component-base/featuregate"
+
+	clusterpediafeature "github.com/clusterpedia-io/clusterpedia/pkg/utils/feature"
+)
+
+const (
+
+	// ApiServerURLRewrite is a feature gate for rewrite apiserver request's URL
+	// owner: @huiwq1990
+	// alpha: v0.7.0
```
Iceber, line 14: Our upcoming release is 0.8.0

pkg/utils/filters/rewrite.go
```diff
@@ -0,0 +1,40 @@
...
+
+func WithRewriteFilter(handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		oldPath := req.URL.EscapedPath()
+		if rewritePath, ok := urlPrefixRewrite(oldPath); ok {
+			req.URL.Path = rewritePath
+			req.Header.Set(OriginPathHeaderKey, oldPath)
+			klog.V(5).InfoS("request need rewrite", "oldPath", oldPath, "newPath", req.URL.EscapedPath())
+		} else {
+			klog.V(5).InfoS("request not need rewrite", "path", oldPath)
```
Iceber, line 23: For paths that don't require rewrite, there seems to be no need to print logs

pkg/utils/filters/rewrite.go
```diff
@@ -0,0 +1,40 @@
...
+		}
+
+		handler.ServeHTTP(w, req)
+	})
+}
+
+func urlPrefixRewrite(oldPath string) (string, bool) {
+	if strings.HasPrefix(oldPath, OldResourceApiServerPrefix) || oldPath == OldResourceApiServerPrefixWithoutSlash {
```
Iceber, line 31: There are more than just `resources` resource under `clusterpedia.io` group, we can filter out resources that start with `apis/clusterpedia.io`.

pkg/utils/filters/rewrite.go
```diff
@@ -0,0 +1,40 @@
...
+		return "", false
+	}
+
+	rewritePath, err := url.JoinPath(OldResourceApiServerPrefix, oldPath)
```
Iceber, line 35: For path like /healthz, we probably don't want to be prefixed.

And since the only prefixes for kube resources are api and apis, we can just rewrite the prefixes for those two.

pkg/apiserver/features/features.go
```diff
@@ -0,0 +1,26 @@
...
+	ApiServerURLRewrite featuregate.Feature = "ApiServerURLRewrite"
```
Iceber, line 15: We may need clearer names, eg. ResourcePathWithoutClusterpediaPrefix

Iceber (commented):
There are some points for discussion

pkg/utils/filters/rewrite.go
huiwq1990: If we process with prefix `apis/clusterpedia.io`, how to process resouce `/apis/clusterpedia.io/v1beta1/clusters` ?

pkg/utils/filters/rewrite.go
huiwq1990: @Iceber In origin implement the /healthz implement by `https://xxx/apis/clusterpedia.io/v1beta1/resources` , so we rewrite it too.

pkg/utils/filters/rewrite.go
Iceber: There is no clusters resource in the `clusterpedia.io` group
pediacluster resource in `clusters.clusterpedia.io` group

pkg/utils/filters/rewrite.go
Iceber: We will provide paths for `/livez`,`/readyz` and other *clusterpedia apiserver application levels* in the future, their paths should not be under resources but under the root path, i don't think it should be required to have only apis/clusterpedia.io/* paths under the root path

pkg/utils/filters/rewrite.go
huiwq1990: Yes, if not support, it will effect apiserice.

Issue #569, opened 2023-09-12
`binding-apiserver` could not delete pods


huiwq1990:
### What happened?

When delete pod in subcluster, binding-apiserver could not delete pods.

binding-apiserver error logs:
![image](https://github.com/clusterpedia-io/clusterpedia/assets/4555057/b51fba3f-e8e8-43fb-a0f1-06ed5c616542)
kubectl --cluster clusterpedia get pod
![image](https://github.com/clusterpedia-io/clusterpedia/assets/4555057/65a2c271-170f-4ab6-9bdb-161b32e7afab)

### What did you expect to happen?

binding-apiserver could delete pods in cache.

### How can we reproduce it (as minimally and precisely as possible)?

Use latest clusterpedia version.

### Anything else we need to know?

I fixed it by modify pkg/synchromanager/clustersynchro/resource_synchro.go which pass real object to delete event.

![image](https://github.com/clusterpedia-io/clusterpedia/assets/4555057/d94f15d7-d337-46ac-b55f-69ac7a154fed)

### Clusterpedia Version

<details>

```console
~ kubectl exec -it -n clusterpedia-system clusterpedia-binding-apiserver-d9dbc5b56-lzrb5 -- binding-apiserver --version
Clusterpedia v0.7.0
```

</details>

### Host Kubernetes version

<details>

```console
~ kubectl version
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.3", GitCommit:"25b4e43193bcda6c7328a6d147b1fb73a33f1598", GitTreeState:"clean", BuildDate:"2023-06-14T09:47:38Z", GoVersion:"go1.20.5", Compiler:"gc", Platform:"darwin/arm64"}
Kustomize Version: v5.0.1
Server Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.3", GitCommit:"25b4e43193bcda6c7328a6d147b1fb73a33f1598", GitTreeState:"clean", BuildDate:"2023-06-15T00:38:14Z", GoVersion:"go1.20.5", Compiler:"gc", Platform:"linux/arm64"}
```

</details>

Iceber:
@wuyingjun-lucky @xyz2277 could you take a look at this, please?

Iceber:
[...]

PR #592, opened 2023-11-15
resource synch process for delete event

huiwq1990:
**What type of PR is this?**

/kind bug

When use binding-apiserver, ondelete should keep object's origin infomation. As memorystorage need object's metadata information.

Fixes https://github.com/clusterpedia-io/clusterpedia/issues/569

```release-note

```

### DIFF
--- a/pkg/synchromanager/clustersynchro/resource_synchro.go
+++ b/pkg/synchromanager/clustersynchro/resource_synchro.go
@@ -317,18 +317,18 @@ func (synchro *ResourceSynchro) OnDelete(obj interface{}) {
 		return
 	}
 
-	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
-	if err != nil {
-		return
+	if d, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+		if obj, ok = d.Obj.(*unstructured.Unstructured); !ok {
+			namespace, name, err := cache.SplitMetaNamespaceKey(d.Key)
+			if err != nil {
+				return
+			}
+			obj = &metav1.PartialObjectMetadata{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: name}}
+		}
 	}
-	namespace, name, err := cache.SplitMetaNamespaceKey(key)
-	if err != nil {
-		return
+	if o, ok := obj.(*unstructured.Unstructured); ok {
+		synchro.pruneObject(o)
 	}
-
-	// Since it is not necessary to save the complete deleted object to the queue,
-	// we convert the object to `PartialObjectMetadata`
-	obj = &metav1.PartialObjectMetadata{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: name}}
 	_ = synchro.queue.Delete(obj)
 }
 


pkg/synchromanager/clustersynchro/resource_synchro.go
```diff
@@ -95,6 +98,8 @@ func newResourceSynchro(cluster string, syncResource schema.GroupVersionResource
 
 		closer: make(chan struct{}),
 		closed: make(chan struct{}),
+
+		transDeleteObject: storage.GetStorageName() != memorystorage.StorageName,
```
Iceber, line 102: Since we support storage layer plugins, it's not great to handle certain logic based on the storage layer name

pkg/synchromanager/clustersynchro/resource_synchro.go
```diff
@@ -317,6 +322,12 @@ func (synchro *ResourceSynchro) OnDelete(obj interface{}) {
 		return
 	}
 
+	if !synchro.transDeleteObject {
+		synchro.pruneObject(obj.(*unstructured.Unstructured))
+		_ = synchro.queue.Delete(obj)
+		return
+	}
```
Iceber, line 329: For the sake of storage generality, it may seem necessary to call Delete with the full object, but there are some special cases where we still need to consider the `DeletedFinalStateUnknown`.

```go

func (synchro *ResourceSynchro) OnDelete(obj interface{}) {
        if !synchro.isRunnableForStorage.Load() {
                return
        }

        if d, ok := obj.(cache.DeletedFinalStateUnknown); ok {
                if obj, ok = d.Obj.(*unstructured.Unstructured); !ok {
                        namespace, name, err := cache.SplitMetaNamespaceKey(d.Key)
                        if err != nil {
                                return
                        }
                        obj = &metav1.PartialObjectMetadata{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: name}}
                }
        }
        if o, ok := obj.(*unstructured.Unstructured); ok {
                synchro.pruneObject(o)
        }
        _ = synchro.queue.Delete(obj)
```

Sometimes, inevitably, we can't get the full object when deleting it.

Iceber (commented):
Thanks for going ahead and fixing this, but there are a few things that still need to be discussed

Iceber:
Are you guys using clusterpedia and would you mind if we communicate via wechat?

huiwq1990:
> Hi @huiwq1990 Are you guys using clusterpedia and would you mind if we communicate via wechat?

ok, I will send my telephone via email.

pkg/synchromanager/clustersynchro/resource_synchro.go
huiwq1990: Yes, as use different storage type, the delete process different. When use database storage, we want to paas only namspace and name, and memory storage we want to paas full object. So we always paas full object both cases?

Iceber: I think we should always pass the full object,  it would be more generic. But sometimes we really can't get the full object and have to pass `PartialObjectMetadata`.

So I think this could be a specification for the storage.Delete interface, which needs to be able to handle the `PartialObjectMetadata` case for the Delete method.

Iceber (approved):
LGTM

But this also requires the memory storage layer to support PartialObjectMetadata, which of course may require other issues and pr to resolve

Also I think that not getting the resource version when deleting may have an effect on https://github.com/clusterpedia-io/clusterpedia/pull/593 as well.

PR #597, opened 2023-11-21
feat: added new parameters for parameterized query

nekomeowww:
## What type of PR is this?

/kind bug

## What this PR does / why we need it

As

https://github.com/clusterpedia-io/clusterpedia/blob/630cf19415ce0148bb1645c7a6f81e686d366051/pkg/storage/internalstorage/util.go#L58-L65

has stated. There are potential SQL injection vulnerabilities of current implementation while users cannot easily defend it due to the lack of support to parameterized query.

This Pull Request 

1. supports parameterized query by using the native feature of [`Where(...)`](https://gorm.io/docs/query.html#Conditions) which GORM supports and will pass the required parameters from `whereSQLParam` and `whereSQLJSONParams` field when 
`whereSQLStatement` was used. Newly added fields:
  - `whereSQLStatement`, drop-in replacement of the existing `whereSQL` field while `whereSQL` will be deprecated in the future release.
  - `whereSQLParam`, for callers of Clusterpedia API to pass the parameters for the raw SQL. Users can send raw queries more safely by using `?` as the placeholder of parameters (which is compatible for GORM) and then include all the corresponding parameters with either one of the newly added field `whereSQLParam` and `whereSQLJSONParams`.
  - `whereSQLJSONParams`, useful when users need more than just `string` type of values of parameters (e.g. `IN` and `NOT IN`).
 2. added a new `AllowParameterizedSQLQuery` feature gate to enable the support of `whereSQLStatement`, `whereSQLParam`, and `whereSQLJSONParams` when transitioning and migrating from `whereSQL` before its deprecation.

For more about the SQL injection of GORM, please read it more here: [https://gorm.io/docs/security.html](https://gorm.io/docs/security.html)

Disclaimer: the potentials of SQL injection will **NEVER** be resolved if implementations of caller still use string concatenation to concat the raw SQL query with parameters directly without `whereSQLParam` or `whereSQLJSONParams` when using `whereSQLStatement`. End users are responsible for defensing and protecting the resources from malicious access at all the time. This Pull Request will only introduce new parameters to help users to defend with parameterized queries.

Besides parameterized query, even though the overall security will be improved with this patch, Clusterpedia cannot guarantee there will be no potential threats of SQL injections in the future and should never be when raw SQL involves. 

End users should check, validate, monitor and restrict the scope of database access as much as possible to reduce the damage and limit the attacking surface even when it failed to defense.

## Which issue(s) this PR fixes

None. This is a known issue.

## Does this PR introduce a user-facing change?

```release-note
**Please update all the existing SQLs of raw queries to pass the parameters of raw SQL in `whereSQLParam` or `whereSQLJSONParams` field instead of string concatenation since it introduces the potential SQL injection when attackers obtained the access to Clusterpedia API.**
```

### DIFF
--- a/pkg/storage/internalstorage/features.go
+++ b/pkg/storage/internalstorage/features.go
@@ -12,6 +12,16 @@ const (
 	// owner: @cleverhu
 	// alpha: v0.3.0
 	AllowRawSQLQuery featuregate.Feature = "AllowRawSQLQuery"
+
+	// AllowParameterizedSQLQuery is a feature gate for the apiserver to allow querying by the parameterized SQL
+	// for better defense against SQL injection.
+	//
+	// Use either single whereSQLStatement field, a pair of whereSQLStatement with whereSQLParam, or
+	// whereSQLStatement with whereSQLJSONParams to pass the SQL it self and parameters.
+	//
+	// owner: @nekomeowww
+	// alpha: v0.8.0
+	AllowParameterizedSQLQuery featuregate.Feature = "AllowParameterizedSQLQuery"
 )
 
 func init() {
@@ -21,5 +31,6 @@ func init() {
 // defaultInternalStorageFeatureGates consists of all known custom internalstorage feature keys.
 // To add a new feature, define a key for it above and add it here.
 var defaultInternalStorageFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
-	AllowRawSQLQuery: {Default: false, PreRelease: featuregate.Alpha},
+	AllowRawSQLQuery:           {Default: false, PreRelease: featuregate.Alpha},
+	AllowParameterizedSQLQuery: {Default: false, PreRelease: featuregate.Alpha},
 }
--- a/pkg/storage/internalstorage/util.go
+++ b/pkg/storage/internalstorage/util.go
@@ -218,7 +218,7 @@ func applyListOptionsToQuery(query *gorm.DB, opts *internal.ListOptions, applyFn
 		query,
 		opts.URLQuery,
 		utilfeature.DefaultMutableFeatureGate.Enabled(AllowRawSQLQuery),
-		false,
+		utilfeature.DefaultMutableFeatureGate.Enabled((AllowParameterizedSQLQuery)),
 	)
 	if err != nil {
 		return 0, nil, nil, err
--- a/pkg/storage/internalstorage/util.go
+++ b/pkg/storage/internalstorage/util.go
@@ -1,7 +1,9 @@
 package internalstorage
 
 import (
+	"encoding/base64"
 	"fmt"
+	"net/url"
 	"strconv"
 	"strings"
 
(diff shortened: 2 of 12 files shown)


pkg/storage/internalstorage/features.go
```diff
@@ -21,5 +21,5 @@ func init() {
 // defaultInternalStorageFeatureGates consists of all known custom internalstorage feature keys.
 // To add a new feature, define a key for it above and add it here.
 var defaultInternalStorageFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
-	AllowRawSQLQuery: {Default: false, PreRelease: featuregate.Alpha},
+	AllowRawSQLQuery: {Default: true, PreRelease: featuregate.Alpha},
```
Iceber, line 24: In general, we couldn't just change the default value of the Alpha feature gate to `true`.
We need to accompany an Alpha upgrade to Beta and create a separate issue to follow the upgrade, but right now it looks like it can't be upgraded to Beta.

pkg/storage/internalstorage/util.go
```diff
@@ -19,9 +20,27 @@ import (
 const (
 	SearchLabelFuzzyName = "internalstorage.clusterpedia.io/fuzzy-name"
 
-	URLQueryWhereSQL = "whereSQL"
+	URLQueryWhereSQL           = "whereSQL"
+	URLQueryWhereSQLParameters = "whereSQLParameters"
```
Iceber, line 24: In order to make it clearer to the user that the sql parameterizes the query, we might use a new query name instead of whereSQL.

`whereSQL` can also be considered deprecated

pkg/storage/internalstorage/util.go
```diff
@@ -19,9 +20,27 @@ import (
...
+
+	// If a string of numbers is passed in from SQL, the query will be taken as ID by default.
+	// If the SQL contains English letter, it will be passed in as column.
+
+	whereParameters, ok := urlQuery[URLQueryWhereSQLParameters]
+	if !ok || len(whereParameters) == 0 {
+		return query.Where(whereClause[0])
+	}
+
+	return query.Where(whereClause[0], whereParameters)
```
Iceber, line 41: need to unpack the params
```go
	return query.Where(whereClause[0], whereParameters...)
```

Iceber: In addition, we need to be able to support the case where the parameters contain arrays, such as a sql statement that contains `IN`, for example `cluster=? AND namespace IN ?`

Iceber (commented):
We could probably add a new feature gate to differentiate from the `AllowRawSQLQuery`/`whereSQL` functionality, this pr would allow us to avoid sql injection if the user is using it properly

The whereSQL will be deprecated (tracked by other issues and prs) and removed for 1.0

Iceber (approved):
LGTM, Thanks for your contribution

PR #609, opened 2023-11-28
add sharding feature for ClusterSynchroManager

### LINKED ISSUE #599: sharding feature for ClusterSynchroManager

Multiple ClusterSynchroManager controllers can synchronize information together by specifying sharding methods

### Why is this needed?

As the number of clusters increases, the CPU and memory resources consumed by ClusterSynchroManager controllers continue to increase, which poses challenges for deploying controllers. Using sharding to allow multiple controllers to synchronize information from different clusters, reducing the number of clusters that each controller needs to synchronize, can effectively reduce the resource consumption of each controller and reduce the difficulty of deploying controllers

jxustc:
**What type of PR is this?**
/kind feature

Implement the basic sharding feature of the ClusterSynchroManager

Fixes #599 

This is a PR for the V1.0 version of the ClusterSynchroManager sharding feature, based on the latest design version discussed previously

```release-note

```

### DIFF
--- a/cmd/clustersynchro-manager/app/options/options.go
+++ b/cmd/clustersynchro-manager/app/options/options.go
@@ -47,6 +47,7 @@ type Options struct {
 
 	WorkerNumber            int // WorkerNumber is the number of worker goroutines
 	PageSizeForResourceSync int64
+	ShardingName            string
 }
 
 func NewClusterSynchroManagerOptions() (*Options, error) {
@@ -90,6 +91,7 @@ func (o *Options) Flags() cliflag.NamedFlagSets {
 	genericfs.Float32Var(&o.ClientConnection.QPS, "kube-api-qps", o.ClientConnection.QPS, "QPS to use while talking with kubernetes apiserver.")
 	genericfs.Int32Var(&o.ClientConnection.Burst, "kube-api-burst", o.ClientConnection.Burst, "Burst to use while talking with kubernetes apiserver.")
 	genericfs.IntVar(&o.WorkerNumber, "worker-number", o.WorkerNumber, "The number of worker goroutines.")
+	genericfs.StringVar(&o.ShardingName, "sharding-name", o.ShardingName, "The sharding name of manager.")
 
 	syncfs := fss.FlagSet("resource sync")
 	syncfs.Int64Var(&o.PageSizeForResourceSync, "page-size", o.PageSizeForResourceSync, "The requested chunk size of initial and resync watch lists for resource sync")
@@ -160,6 +162,10 @@ func (o *Options) Config() (*config.Config, error) {
 	}
 	kubeStateMetricsServerConfig := o.KubeStateMetrics.ServerConfig(metricsConfig)
 
+	if o.ShardingName != "" {
+		o.LeaderElection.ResourceName = fmt.Sprintf("%s-%s", o.LeaderElection.ResourceName, o.ShardingName)
+	}
+
 	return &config.Config{
 		CRDClient:     crdclient,
 		Kubeconfig:    kubeconfig,
@@ -167,6 +173,7 @@ func (o *Options) Config() (*config.Config, error) {
 
 		StorageFactory: storagefactory,
 		WorkerNumber:   o.WorkerNumber,
+		ShardingName:   o.ShardingName,
 
 		MetricsServerConfig:     metricsConfig,
 		KubeMetricsServerConfig: kubeStateMetricsServerConfig,
--- a/pkg/synchromanager/clustersynchro_manager.go
+++ b/pkg/synchromanager/clustersynchro_manager.go
@@ -48,6 +48,7 @@ type Manager struct {
 	clusterpediaclient crdclientset.Interface
 	informerFactory    externalversions.SharedInformerFactory
 
+	shardingName               string
 	queue                      workqueue.RateLimitingInterface
 	storage                    storage.StorageFactory
 	clusterlister              clusterlister.PediaClusterLister
@@ -62,7 +63,7 @@ type Manager struct {
 
 var _ kubestatemetrics.ClusterMetricsWriterListGetter = &Manager{}
 
-func NewManager(client crdclientset.Interface, storage storage.StorageFactory, syncConfig clustersynchro.ClusterSyncConfig) *Manager {
+func NewManager(client crdclientset.Interface, storage storage.StorageFactory, syncConfig clustersynchro.ClusterSyncConfig, shardingName string) *Manager {
 	factory := externalversions.NewSharedInformerFactory(client, 0)
 	clusterinformer := factory.Cluster().V1alpha2().PediaClusters()
 	clusterSyncResourcesInformer := factory.Cluster().V1alpha2().ClusterSyncResources()
@@ -70,6 +71,7 @@ func NewManager(client crdclientset.Interface, storage storage.StorageFactory, s
 	manager := &Manager{
 		informerFactory:    factory,
 		clusterpediaclient: client,
+		shardingName:       shardingName,
 
 		storage:                    storage,
 		clusterlister:              clusterinformer.Lister(),
@@ -166,7 +168,9 @@ func (manager *Manager) addCluster(obj interface{}) {
 func (manager *Manager) updateCluster(older, newer interface{}) {
 	oldObj := older.(*clusterv1alpha2.PediaCluster)
 	newObj := newer.(*clusterv1alpha2.PediaCluster)
-	if newObj.DeletionTimestamp.IsZero() && equality.Semantic.DeepEqual(oldObj.Spec, newObj.Spec) {
+	if newObj.DeletionTimestamp.IsZero() &&
+		equality.Semantic.DeepEqual(oldObj.Spec, newObj.Spec) &&
+		oldObj.Status.ShardingName == newObj.Status.ShardingName {
 		return
 	}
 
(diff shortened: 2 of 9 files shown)


pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -183,6 +188,12 @@ func (manager *Manager) enqueue(obj interface{}) {
 		return
 	}
 
+	cluster := obj.(*clusterv1alpha2.PediaCluster)
+	if manager.shardingName != cluster.Spec.ShardingName &&
+		(cluster.Status.ShardingName == nil || manager.shardingName != *cluster.Status.ShardingName) {
+		return
+	}
```
Iceber, line 195: Missing filter `spec.sharding == name and status.sharding != nil and status.sharding != name`

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -286,10 +297,23 @@ func (manager *Manager) reconcileCluster(cluster *clusterv1alpha2.PediaCluster)
 		}
 	}
 
+	if cluster.Spec.ShardingName == manager.shardingName {
+		if cluster.Status.ShardingName != nil && *cluster.Status.ShardingName != manager.shardingName {
```
Iceber, line 301: Judging this case here would result in an error deleting the cluster for which `status.sharding ! = name`, as there may be concurrent processing of the same cluster with other synchronization managers.

If a cluster is deleted by another cluster before the manager corresponding to status.sharding has started processing the cluster to be deleted, the cluster synchronizer may not be able to shut down, and dirty data may occur.

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -286,10 +297,23 @@ func (manager *Manager) reconcileCluster(cluster *clusterv1alpha2.PediaCluster)
...
+			return controller.NoRequeueResult
+		}
+	}
+
 	manager.synchrolock.RLock()
 	synchro := manager.synchros[cluster.Name]
 	manager.synchrolock.RUnlock()
 
+	if synchro != nil {
```
Iceber, line 310: When the conditions are met it should be unmanaged with or without synchro.

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -398,6 +428,23 @@ func (manager *Manager) reconcileCluster(cluster *clusterv1alpha2.PediaCluster)
 	return controller.NoRequeueResult
 }
 
+func (manager *Manager) reconcileShardingUnmanaged(cluster *clusterv1alpha2.PediaCluster) controller.Result {
+	if *cluster.Status.ShardingName == manager.shardingName {
+		if err := manager.removeCluster(cluster.Name); err != nil {
```
Iceber, line 433: removeCluster should not be called, it will clean up the data in the storage tier, we just need to stop synchro, we don't need to delete the data.

Iceber:
Rearranged the logic（pseudocode）：
```
func reconcileCluster (){
   // To ensure coherent logic, we can add prejudgements in reconcileCluster
    if status.sharding == nil && spec.sharding != name{
        return
    }

    if status.sharding != nil && status.sharding != name {
        return
    }

    // The cluster will be in the following state：
    // spec.sharding == name and (status.sharding ==nil or status.sharding == name)
    // spec.sharding != name and status != nil and status.sharding == name


    if delele {
        do_delete....
        return
    }

    if spec.sharding != name {
        unmanage_and_set_nil()
        return
    }

    set_sharding_name()
    ...
}
```
It might be useful.

jxustc:
> Rearranged the logic（pseudocode）：
> 
> ```
> func reconcileCluster (){
>    // To ensure coherent logic, we can add prejudgements in reconcileCluster
>     if status.sharding == nil && spec.sharding != name{
>         return
>     }
> 
>     if status.sharding != nil && status.sharding != name {
>         return
>     }
> 
>     // The cluster will be in the following state：
>     // spec.sharding == name and (status.sharding ==nil or status.sharding == name)
>     // spec.sharding != name and status != nil and status.sharding == name
> 
> 
>     if delele {
>         do_delete....
>         return
>     }
> 
>     if spec.sharding != name {
>         unmanage_and_set_nil()
>         return
>     }
> 
>     set_sharding_name()
>     ...
> }
> ```
> 
> It might be useful.

Thanks for the suggestion, I will improve it over the weekend

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -286,6 +312,10 @@ func (manager *Manager) reconcileCluster(cluster *clusterv1alpha2.PediaCluster)
 		}
 	}
 
+	if cluster.Spec.ShardingName != manager.shardingName {
+		return manager.reconcileShardingUnmanaged(cluster)
+	}
+
```
Iceber, line 318: We can set `status.ShardingName` to `manager.shardingName` in time here, and update it to indicate that the PediaCluster has been worked on by this manager before the manager updates the rest of the PediaCluster's status. Otherwise, the shardingName might stay nil, but the other status might keep updating due to an anomaly

cmd/clustersynchro-manager/app/options/options.go
```diff
@@ -90,6 +91,10 @@ func (o *Options) Flags() cliflag.NamedFlagSets {
 	genericfs.Float32Var(&o.ClientConnection.QPS, "kube-api-qps", o.ClientConnection.QPS, "QPS to use while talking with kubernetes apiserver.")
 	genericfs.Int32Var(&o.ClientConnection.Burst, "kube-api-burst", o.ClientConnection.Burst, "Burst to use while talking with kubernetes apiserver.")
 	genericfs.IntVar(&o.WorkerNumber, "worker-number", o.WorkerNumber, "The number of worker goroutines.")
+	genericfs.StringVar(&o.ShardingName, "sharding-name", o.ShardingName, "The name of the sharding that the manager is to synchro.")
+	if o.ShardingName != "" {
```
Iceber, line 95: o.shardingName must be "" when Flags() is called, so this logic won't be handled

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -532,6 +591,12 @@ func (manager *Manager) updateClusterStatus(ctx context.Context, name string, up
 	})
 }
 
+func (manager *Manager) UpdateClusterShardingStatus(ctx context.Context, name string, status *clusterv1alpha2.ClusterStatus) error {
```
Iceber, line 594: Changing the parameter to `shardingName` makes it clearer.
`func (manager *Manager) UpdateClusterShardingStatus(ctx context.Context, name string, shardingName *string)`

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -286,6 +312,22 @@ func (manager *Manager) reconcileCluster(cluster *clusterv1alpha2.PediaCluster)
 		}
 	}
 
+	if cluster.Spec.ShardingName != manager.shardingName {
+		if *cluster.Status.ShardingName == manager.shardingName {
```
Iceber, line 316: By the previous rule, `status.ShardingName` must be equal to `manager.shardingName` when `cluster.Spec.ShardingName != manager.shardingName`

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -256,6 +271,17 @@ func (manager *Manager) processNextCluster() (continued bool) {
 
 // if err returned is not nil, cluster will be requeued
 func (manager *Manager) reconcileCluster(cluster *clusterv1alpha2.PediaCluster) controller.Result {
+	if cluster.Status.ShardingName == nil && cluster.Spec.ShardingName != manager.shardingName {
+		return controller.NoRequeueResult
+	}
+
+	if cluster.Status.ShardingName != nil && *cluster.Status.ShardingName != manager.shardingName {
+		return controller.NoRequeueResult
+	}
```
Iceber, line 280: I think this could be put into a separate function, which would add more comprehensive unit testing to it.
But it can be optimized in a new pr.

pkg/synchromanager/clustersynchro_manager.go
```diff
@@ -398,6 +446,17 @@ func (manager *Manager) reconcileCluster(cluster *clusterv1alpha2.PediaCluster)
...
 }
 
+func (manager *Manager) stopClusterSynchro(name string) {
+	manager.synchrolock.Lock()
+	synchro := manager.synchros[name]
+	delete(manager.synchros, name)
+	manager.synchrolock.Unlock()
+
+	if synchro != nil {
+		synchro.Shutdown(false)
```
Iceber, line 456: Here you need to use true, to trigger a status update for the pediacluster.

This is not the same as `removeCluster`.

cmd/clustersynchro-manager/app/options/options.go
```diff
@@ -90,6 +91,7 @@ func (o *Options) Flags() cliflag.NamedFlagSets {
 	genericfs.Float32Var(&o.ClientConnection.QPS, "kube-api-qps", o.ClientConnection.QPS, "QPS to use while talking with kubernetes apiserver.")
 	genericfs.Int32Var(&o.ClientConnection.Burst, "kube-api-burst", o.ClientConnection.Burst, "Burst to use while talking with kubernetes apiserver.")
 	genericfs.IntVar(&o.WorkerNumber, "worker-number", o.WorkerNumber, "The number of worker goroutines.")
+	genericfs.StringVar(&o.ShardingName, "sharding-name", o.ShardingName, "The name of the sharding that the manager is to synchro.")
```
Iceber, line 94: This can be simplified to `The sharding name of manager`

Iceber (commented):
Others look good, only a little bit of modification needed

cmd/clustersynchro-manager/app/options/options.go
```diff
@@ -160,13 +162,18 @@ func (o *Options) Config() (*config.Config, error) {
 	}
 	kubeStateMetricsServerConfig := o.KubeStateMetrics.ServerConfig(metricsConfig)
 
+	if o.ShardingName != "" {
+		o.LeaderElection.ResourceName = fmt.Sprintf("%s-%s", o.LeaderElection.ResourceName, o.ShardingName)
```
calvin0327, line 166: Do we need to check whether  the resource already exists here?

Iceber: The leaderelection will take care of checking for the resource, and if it already exists then it will race for it.

Iceber (approved):
LGTM, Thanks

PR #615, opened 2023-12-08
add featureGate IgnoreLeaseSync to stop sync leases.coordination.k8s.io

27149chen:
**What type of PR is this?**

/kind feature

This pr adds a featureGate `IgnoreLeaseSync` to stop synchronizing internal resources that the majority of users do not care about. Currently only `leases.coordination.k8s.io` is added since it updates too frequently.

Fixes #

```release-note
Add a featureGate `SkipSyncLeases` to stop synchronizing `leases.coordination.k8s.io`
```

### DIFF
--- a/pkg/synchromanager/clustersynchro/resource_negotiator.go
+++ b/pkg/synchromanager/clustersynchro/resource_negotiator.go
@@ -87,6 +87,14 @@ func (negotiator *ResourceNegotiator) NegotiateSyncResources(syncResources []clu
 	for _, groupResources := range syncResources {
 		for _, resource := range groupResources.Resources {
 			syncGR := schema.GroupResource{Group: groupResources.Group, Resource: resource}
+
+			if clusterpediafeature.FeatureGate.Enabled(features.IgnoreLeaseSync) {
+				// skip leases.coordination.k8s.io
+				if syncGR.String() == "leases.coordination.k8s.io" {
+					continue
+				}
+			}
+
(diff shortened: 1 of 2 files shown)


pkg/synchromanager/clustersynchro/resource_negotiator.go
```diff
@@ -21,6 +21,12 @@ import (
 	clusterpediafeature "github.com/clusterpedia-io/clusterpedia/pkg/utils/feature"
 )
 
+// excludedInternalResources is a list of internal resources that are not synchronized
+// if the feature gate SkipSyncLeases is enabled.
+var excludedInternalResources = sets.NewString(
+	"leases.coordination.k8s.io",
+)
```
Iceber, line 28: We've made feature gates specific to the leases resource, so I don't think we need to maintain the `excludedInternalResources` here, we could add it if we need it more in the future

Iceber (commented):
The rest looks good

Iceber:
If there is more demand for skipping resource configurations in the future, then it may be necessary to add a `--exclude-resources flag`, but I don't think I'd really want to do that, because the people who use clusterpedia and the people who deploy clusterpedia may not be the same, and it's not appropriate to silently skip some resources.

In the future it is still recommended to exclude resources on PediaCluster.

This pr allows users who have created a large number of pediaclusters to skip leases without modifying PediaCluster, which is a good value for existing users.

Iceber (approved):
LGTM

PR #638, opened 2024-01-05
fix otel trace handler WithTracing, use another func

KubeKyrie:
**What type of PR is this?**
/kind bug

Fixed OTEL trace interruption when other services requested clusterpedia apiserver. And this is caused by `otelhttp.WithPublicEndpoint()`. As shown here https://github.com/kubernetes/kubernetes/blob/09a5049ca785024edd4955eb82e855d9b5657491/staging/src/k8s.io/apiserver/pkg/endpoints/filters/traces.go#L33C20-L33C20

And I think it's a bug of kubernetes, so I  replaced it with `WithTracing` in pkg k8s.io/component-base/tracing.

Fixes #None

The two `WithTracing` is almost same and we can customize the span name by the new one. So it's more flexible.

```release-note
fix otel trace handler WithTracing
```

### DIFF
--- a/pkg/kubeapiserver/apiserver.go
+++ b/pkg/kubeapiserver/apiserver.go
@@ -17,6 +17,7 @@ import (
 	"k8s.io/apiserver/pkg/server/healthz"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/client-go/restmapper"
+	"k8s.io/component-base/tracing"
 
 	informers "github.com/clusterpedia-io/clusterpedia/pkg/generated/informers/externalversions"
 	"github.com/clusterpedia-io/clusterpedia/pkg/kubeapiserver/discovery"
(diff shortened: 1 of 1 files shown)


KubeKyrie:
The [otelhttp.WithPublicEndpoint()](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/6d226aff737e650125fd98e18ccba66428f328f7/instrumentation/net/http/otelhttp/config.go#L103) option causes spans generated by the API Server handler to be the start of a new trace (with a link to the incoming trace), rather than be children of the incoming request.
The reason why kube-apiserver uses WithPublicEndpoint is to prevent callers from influencing sampling decisions, see detail https://github.com/kubernetes/kubernetes/issues/109829. And the discussion about the reentrant problem is still in progress, see https://github.com/kubernetes/kubernetes/issues/103186. 

So, I think the current solution is to use another handler filter func, just as this PR shows.

Iceber (approved):
LGTM

PR #639, opened 2024-01-07
add image build dependency

KubeKyrie:
**What type of PR is this?**
/kind feature

Fixes #None

Add image build dependency so that we can directly `make image-apiserver` when builder not exists.

```release-note
Add image build dependency
```


Iceber:
you could use `make build-image-apiserver`
https://github.com/clusterpedia-io/clusterpedia/blob/b4117dd63e0a7aebd944b7b817bfea28ebd7c447/Makefile#L95-L100

PR #741, opened 2025-03-05
apiserver: forward request with prefix path or header

Iceber:
**What type of PR is this?**

/kind feature

There are two ways to proxy requests to member clusters
**prefix path**
```yaml
- cluster:
    insecure-skip-tls-verify: true
    server: https://127.0.0.1:8443/apis/clusterpedia.io/v1beta1/resources/clusters/cluster-example/
  name: cluster-example
- cluster:
    insecure-skip-tls-verify: true
    server: https://127.0.0.1:8443/apis/clusterpedia.io/v1beta1/resources/clusters/cluster-example/proxy
  name: cluster-example-proxy
```
To enable this functionality, append the "/proxy" suffix to the server endpoint in your kubeconfig file. The cluster can then be used normally.
```bash
kubectl --cluster cluster-example-proxy get po -A
```

**Header `X-CLUSTERPEDIA-FORWARD`**
```bash
curl -v -k --cert-type P12 --cert client.p12:password  -H "X-CLUSTERPEDIA-FORWARD: true" "https://localhost:8443/apis/clusterpedia.io/v1beta1/resources/clusters/cluster-example/api/v1/namespaces/clusterpedia-system/pods"
```
Note that the alternative approach of passing this information via HTTP headers is not compatible with kubectl commands.

Fixes #

```release-note

```

### DIFF
--- a/pkg/kubeapiserver/apiserver.go
+++ b/pkg/kubeapiserver/apiserver.go
@@ -130,30 +130,15 @@ func (c completedConfig) New(delegationTarget genericapiserver.DelegationTarget)
 
 	delegate := delegationTarget.UnprotectedHandler()
 	if delegate == nil {
+		// To prevent anomalous requests from being sent to member clusters due to improper usage,
+		// we do not allow proxying requests with unmatched prefixes to member clusters.
 		delegate = http.NotFoundHandler()
 	}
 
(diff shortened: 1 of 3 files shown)


Iceber:
// Prior to the new version release(0.9), we should discuss and implement strict flags and feature gates

PR #830, opened 2026-04-09
kubeapiserver: normalize pod portforward proxy subresource option

### LINKED ISSUE #829: kubeapiserver: pods/portforward cannot be enabled via --allowed-proxy-subresources
### What happened?

`pkg/kubeapiserver/options.go` validates `--allowed-proxy-subresources` against `supportedProxyCoreSubresources`, but the pod subresource list currently contains `portfowrd` instead of `portforward`.

That creates two broken behaviors in the current code:

1. `--allowed-proxy-subresources=pods/portforward` is rejected as unsupported.
2. `--allowed-proxy-subresources=pods/portfowrd` is accepted, but it still does not enable pod port-forward proxying because the actual registered subresource name in `pkg/kubeapiserver/resourcerest/proxy/subresource.go` is `portforward`.

In practice, pod port-forward proxy requests cannot be enabled through the documented flag path.

### What did you expect to happen?

`pods/portforward` should be accepted by `--allowed-proxy-subresources` and should enable the registered pod `portforward` subresource.

For compatibility, the historical typo should either be normalized to `portforward` or otherwise handled without leaving the feature unusable.

### How can we reproduce it (as minimally and precisely as possible)?

1. Check the current allowlist in `pkg/kubeapiserver/options.go`.
2. Check the registered proxy subresources in `pkg/kubeapiserver/resourcerest/proxy/subresource.go`.
3. Start the apiserver with `--feature-gates=AllowProxyRequestToClusters=true --allowed-proxy-subresources=pods/portforward`.
4. Observe that validation rejects `pods/portforward` as unsupported.
5. Start the apiserver with `--feature-gates=AllowProxyRequestToClusters=true --allowed-proxy-subresources=pods/portfowrd`.
6. Observe that the typo is accepted by validation, but the apiserver still never registers the real `pods/portforward` proxy subresource because the allowlist entry does not match the actual REST subresource name.

### Anything else we need to know?

This is visible directly in `main` at commit `92e3fafb`.

The mismatch is between:

- `pkg/kubeapiserver/options.go`: `portfowrd`
- `pkg/kubeapiserver/resourcerest/proxy/subresource.go`: `portforward`

### Clusterpedia Version

<details>

```console
N/A for a live deployment.
Observed directly in the current main branch source at commit 92e3fafb.
```

</details>

### Host Kubernetes version

<details>

```console
N/A.
This is a flag-validation and subresource-registration bug in the current codebase.
```

</details>

officialasishkumar:
**What type of PR is this?**

/kind bug

The proxy REST layer registers `pods/portforward`, but `--allowed-proxy-subresources` only advertised and validated the misspelled `portfowrd` token. That left the feature unusable through the flag path: the correct spelling was rejected, while the accepted typo never matched the actual registered subresource.

This change makes `portforward` the canonical pod subresource name, keeps `portfowrd` as a compatibility alias, and adds unit tests around the parser.

Fixes #829

Validated with:
- `go test ./pkg/kubeapiserver/... -count=1`
- `go test ./pkg/... -count=1`

```release-note
Fixed `--allowed-proxy-subresources` so pod `portforward` proxying can be enabled with `pods/portforward`, while the historical `pods/portfowrd` typo remains accepted as a compatibility alias.
```

### DIFF
--- a/pkg/kubeapiserver/options.go
+++ b/pkg/kubeapiserver/options.go
@@ -62,29 +62,39 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 }
 
 var supportedProxyCoreSubresources = map[string][]string{
-	"pods":     {"proxy", "log", "exec", "attach", "portfowrd"},
+	"pods":     {"proxy", "log", "exec", "attach", "portforward"},
 	"nodes":    {"proxy"},
 	"services": {"proxy"},
 }
 
-func (o *Options) Config() (*ExtraConfig, error) {
-	if !utilfeature.DefaultFeatureGate.Enabled(features.AllowProxyRequestToClusters) && (len(o.AllowedProxySubresources) != 0 ||
-		o.EnableProxyPathForForwardRequest || o.AllowForwardUnsyncResourceRequest) {
-		return nil, fmt.Errorf("please enable feature gate %s to allow apiserver to handle the proxy and forward requests", features.AllowProxyRequestToClusters)
-	}
+// TODO(Iceber): remove the legacy portfowrd alias in 1.0.
+var proxySubresourceAliases = map[string]string{
+	"portfowrd": "portforward",
+}
 
+func normalizeProxySubresourceName(name string) string {
+	if canonical, ok := proxySubresourceAliases[name]; ok {
+		return canonical
+	}
+	return name
+}
+
+func parseAllowedProxySubresources(allowed []string) (map[schema.GroupResource]sets.Set[string], error) {
 	subresources := make(map[schema.GroupResource]sets.Set[string])
 
-	for _, subresource := range o.AllowedProxySubresources {
+	for _, subresource := range allowed {
+		raw := strings.TrimSpace(subresource)
 		var resource string
-		switch slice := strings.Split(strings.TrimSpace(subresource), "/"); len(slice) {
+		switch slice := strings.Split(raw, "/"); len(slice) {
 		case 1:
 			subresource = slice[0]
 		case 2:
 			resource, subresource = slice[0], slice[1]
 		default:
-			return nil, fmt.Errorf("--allowed-proxy-subresources: invalid format %q", subresource)
+			return nil, fmt.Errorf("--allowed-proxy-subresources: invalid format %q", raw)
 		}
+
+		subresource = normalizeProxySubresourceName(subresource)
 
 		var matched bool
 		for r, srs := range supportedProxyCoreSubresources {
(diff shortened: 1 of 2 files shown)


pkg/kubeapiserver/options.go
```diff
@@ -106,6 +114,19 @@ func (o *Options) Config() (*ExtraConfig, error) {
 			return nil, fmt.Errorf("--allowed-proxy-subresources: unsupported subresources or invalid format %q", subresource)
 		}
 	}
```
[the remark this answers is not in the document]
officialasishkumar: Handled in c43a1443 by preserving the trimmed raw token for both parser error paths and covering those cases in tests.

pkg/kubeapiserver/options.go
```diff
@@ -62,30 +62,39 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
...
+	"pods":     {"proxy", "log", "exec", "attach", "portforward"},
 	"nodes":    {"proxy"},
 	"services": {"proxy"},
 }
 
-func (o *Options) Config() (*ExtraConfig, error) {
-	if !utilfeature.DefaultFeatureGate.Enabled(features.AllowProxyRequestToClusters) && (len(o.AllowedProxySubresources) != 0 ||
-		o.EnableProxyPathForForwardRequest || o.AllowForwardUnsyncResourceRequest) {
-		return nil, fmt.Errorf("please enable feature gate %s to allow apiserver to handle the proxy and forward requests", features.AllowProxyRequestToClusters)
+var proxySubresourceAliases = map[string]string{
```
Iceber, line 71: For compatibility, we do need to use the alias approach, but we can remove it in a future 1.0 release. We can add a TODO to remind ourselves — for example, `TODO: remove in 1.0`.

Iceber (commented):
LGTM, Thanks for catching this bug. Other than that, LGTM.

Iceber:
Please squash the commits into one, ensuring that related and minor changes are contained within a single commit.

pkg/kubeapiserver/options.go
officialasishkumar: Added the TODO for removing the compatibility alias in 1.0 and force-pushed the branch.

.github/workflows/ci.yml
```diff
@@ -87,7 +87,7 @@ jobs:
         run: ./test/test.sh
 
       - name: Upload logs
-        uses: actions/upload-artifact@v7
+        uses: actions/upload-artifact@v6
```
Iceber: @officialasishkumar Here's an unexpected rollback.

officialasishkumar: Restored `actions/upload-artifact@v7`.

Iceber (approved):

PR #859, opened 2026-05-11
storage: reject negative owner seniority

### LINKED ISSUE #858: Negative ownerSeniority can recurse indefinitely in owner queries
### What happened?

List requests can pass a negative owner seniority through the `ownerSeniority` URL parameter. The value is parsed as a signed integer and is not rejected by Clusterpedia's list-option conversion.

That value later reaches the internal storage owner-query builders. Both recursive builders decrement `seniority` until it is exactly `0`; a negative value never reaches that base case, so a request such as `?ownerUID=<uid>&ownerSeniority=-1` can recurse indefinitely and exhaust the apiserver goroutine stack instead of returning a validation error.

The affected current-main code paths are:

- `staging/src/github.com/clusterpedia-io/api/clusterpedia/v1beta1/conversion.go`, where the signed URL value is accepted.
- `pkg/storage/internalstorage/json_builder.go`, where `buildOwnerQueryByUID` and `buildOwnerQueryByName` recurse by subtracting one.

### What did you expect to happen?

Clusterpedia should reject negative `ownerSeniority` values with a bad request / invalid list-options error before storage query construction. Programmatic storage calls should also avoid entering the recursive owner-query builders with a negative value.

### How can we reproduce it (as minimally and precisely as possible)?

Against current `main`, send an owner search request with a negative seniority for a single cluster, for example:

```console
kubectl --cluster <member-cluster> get pods -A --raw '/api/v1/pods?ownerUID=<uid>&ownerSeniority=-1'
```

The request decoding path accepts the value, and the internal storage owner-query builder has no terminating condition for negative seniority.

### Anything else we need to know?

This is a request validation bug and does not depend on a particular storage backend beyond using the internal storage owner search path.

### Clusterpedia Version

Current `main` at commit `bece343b`.

### Host Kubernetes version

Not Kubernetes-version specific.

officialasishkumar:
**What type of PR is this?**

/kind bug

Rejects negative `ownerSeniority` values during v1beta1 list-option conversion so invalid owner-search requests fail before reaching storage query construction.

Adds a defensive internal storage validation before recursive owner query builders are called, so programmatic callers cannot trigger unbounded recursion with a negative seniority value.

Fixes #858

NONE

```release-note
Reject negative ownerSeniority list queries instead of allowing owner search query construction to recurse indefinitely.
```

### DIFF
--- a/pkg/storage/internalstorage/resource_storage.go
+++ b/pkg/storage/internalstorage/resource_storage.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/apimachinery/pkg/watch"
 	genericstorage "k8s.io/apiserver/pkg/storage"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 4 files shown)


Iceber (approved):
LGTM, Thank you for your contribution!

 Sorry for the delay in reviewing this PR.

PR #718, opened 2025-01-02
clustersynchro: add resource synchro metrics

Iceber:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

Issue #699, opened 2024-09-12
Add database operation metrics


huiwq1990:
### What would you like to be added?

Expose metrics url for `clustersynchro-manager`, and monitor database operations. Just like: https://github.com/profects/gormetrics

### Why is this needed?

We want to know the pressure on the database caused by clustersynchro-manager component.

PR #867, opened 2026-06-23
internalstorage: add database operation metrics

mayur-tolexo:
Addresses #699.

The storage layer already exposes connection-pool metrics (`storage_dbstats_*`), but nothing about the database operations themselves, so it's hard to see the load `clustersynchro-manager` and the apiserver put on the database. This adds that.

I added a GORM plugin that hooks the create/query/update/delete/row/raw callbacks (same way `tracing.go` does) and records two metrics per operation:

- `storage_operation_total{operation,status}` — count, with `status` either `success` or `error`
- `storage_operation_duration_seconds{operation}` — latency

A "record not found" (and the other expected no-rows outcomes the tracing plugin already ignores) is counted as success rather than an error. The plugin carries the same `db_name` const label as the existing DBStats metrics, and is registered next to them under the same `metrics.disable` switch, so there's no new config.

I kept this separate from the resource-synchro metrics added in #718 — those count resource operations at the sync layer, whereas this is about actual database load.

Tested with a unit test that drives the callbacks against a sqlite DB and checks the success/error counters and the latency histogram. `go build ./...`, `go vet` and the package tests pass.

I'm happy to adjust the metric names/labels or the approach if you'd prefer something different.

### DIFF
--- a/pkg/storage/internalstorage/operation_metrics.go
+++ b/pkg/storage/internalstorage/operation_metrics.go
@@ -1,0 +1,132 @@
+package internalstorage
+
+import (
+	"database/sql"
+	"database/sql/driver"
+	"errors"
+	"fmt"
+	"io"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"gorm.io/gorm"
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
+// opStartTimeKey is the GORM instance setting that carries an operation's start
+// time from its before callback to its after callback.
+const opStartTimeKey = "clusterpedia:operation_start"
+
+var _ gorm.Plugin = &operationMetrics{}
+
+// operationMetrics is a gorm.Plugin that records the count and latency of
+// database operations (create/query/update/delete/row/raw) via GORM callbacks.
+// It complements the connection-pool metrics provided by Prometheus.
+type operationMetrics struct {
+	dbName   string
+	total    *prometheus.CounterVec
+	duration *prometheus.HistogramVec
+}
+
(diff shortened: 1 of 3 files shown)


mayur-tolexo:
Quick sanity check — ran one of each operation against a sqlite DB and looked at the collectors. Counters come out right:

```
storage_operation_total{db_name="clusterpedia",operation="create",status="success"} 1
storage_operation_total{db_name="clusterpedia",operation="delete",status="success"} 1
storage_operation_total{db_name="clusterpedia",operation="query",status="error"} 1
storage_operation_total{db_name="clusterpedia",operation="query",status="success"} 2
storage_operation_total{db_name="clusterpedia",operation="raw",status="success"} 1
storage_operation_total{db_name="clusterpedia",operation="update",status="success"} 1
```

The error one is from querying a table that doesn't exist, query shows 2 from two Finds, and raw is an `Exec("SELECT 1")`.

Latencies are recorded too (query count is 3 — the failed query is still timed, since the callbacks run whether or not it errors):

```
storage_operation_duration_seconds_count{...,operation="create"} 1
storage_operation_duration_seconds_count{...,operation="query"} 3
storage_operation_duration_seconds_count{...,operation="update"} 1
storage_operation_duration_seconds_count{...,operation="delete"} 1
storage_operation_duration_seconds_count{...,operation="raw"} 1
```

And the unit test passes:

```
$ go test ./pkg/storage/internalstorage/ -run TestOperationMetrics
ok  github.com/clusterpedia-io/clusterpedia/pkg/storage/internalstorage
```

Iceber (approved):
LGTM, Thank you for your contribution！

Sorry for the delay in reviewing this PR.

